Issue #18, opened 2021-11-16
Add search endpoint


jswk:
See https://github.com/cyfronet-fid/eosc-search-service-api/blob/4cd59a96dc6ac0ce2d2046359770beb0af9409c3/eosc-search-service-v1.yaml#L12 for possible definition.
The search query should be passed on to the search engine, paging from the search engine should be exposed as well (rows, cursor).

The endpoint definition lacks filters, add them. Use [filter query](https://nightlies.apache.org/solr/draft-guides/solr-reference-guide-main/common-query-parameters.html#fq-filter-query-parameter) to pass them.

PR #28, opened 2022-01-20
Add query fields and filter query to search params

jswk:
Query fields (qf) is necessary in order to be able to query a custom subset
of documents. Otherwise the default Solr schema searches a common _text_ field.
But this has several drawbacks:
- the index is twice the size, since all fields must be copied there,
- query value transformations (e.g. n-grams) cannot be applied directly on a selected field,
- the choice on what fields to search is done when designing a schema, not when
  calling the ESS.

Add also filter query (fq) to filter results by specific fields, apart from filtering by q.

Closes: #18.

### DIFF
--- a/backend/app/recommender/operations.py
+++ b/backend/app/recommender/operations.py
@@ -1,0 +1,50 @@
+"""Operations on Recommender Service"""
+from httpx import AsyncClient, Response
+
+from app.config import RS_URL
+from app.requests import RecommendRequest
+
+
+async def recommendations(
+    client: AsyncClient,
+    collection: str,
+    candidates: list[hash],
+    *,
+    context: RecommendRequest,
+    q: str,
+    qf: list[str],
+    fq: list[str],
+    sort: list[str],
+) -> Response:
+    """
+    Retrieve recommendations from RS based on candidates.
+
+    The q, qf, fq, sort are passed to RS as search_data.
+    See https://solr.apache.org/guide/8_11/query-syntax-and-parsing.html subsections for their
+    meaning, they may not be used directly by the RS though.
+    """
+    payload = {
+        "unique_id": context.unique_id,
+        "timestamp": str(context.timestamp),
+        "visit_id": context.visit_id,
+        "page_id": context.page_id,
+        "panel_id": [context.panel_id],
+        "num_lists": 1,
+        "candidates": [doc_to_candidate(doc) for doc in candidates],
+        "search_data": {
+            "q": q,
+            "sort": ", ".join(sort),
+            "collection": collection,
+            "qf": qf,
+            "fq": fq,
+        },
+    }
+    return await client.post(f"{RS_URL}recommendations", json=payload)
+
+
+def doc_to_candidate(doc):
+    """Convert a document from Solr response to its first pid or id"""
+    try:
+        return doc["pid"][0]
+    except (KeyError, IndexError):
+        return doc["id"]
--- a/backend/app/routes/recommend.py
+++ b/backend/app/routes/recommend.py
@@ -1,30 +1,28 @@
 """The Recommend endpoint"""
-from datetime import datetime
 from json import JSONDecodeError
 
 from fastapi import Body, HTTPException, Query
 from httpx import AsyncClient
-from pydantic import BaseModel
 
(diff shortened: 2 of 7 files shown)


backend/app/recommender/operations.py
```diff
@@ -0,0 +1,44 @@
...
+from app.requests import RecommendRequest
+
+
+async def recommendations(
+    client: AsyncClient,
+    collection: str,
+    candidates: list[hash],
+    *,
+    context: RecommendRequest,
+    q: str,
```
danielkryska, line 14: Could you describe params or add a URL to its documentation? Letters q, qf, and fq don't mean anything to me. What value accept collection, candidates, and sort?

danielkryska: Could you also provide in function few examples of usage?

danielkryska: The best possible example would be 1 parameter change per example.

backend/app/routes/recommend.py
```diff
@@ -1,30 +1,28 @@
...
-    page_id: str
-    panel_id: str
-
-
+# pylint: disable=too-many-arguments
 @internal_api_router.post("/recommend")
 async def recommend_post(
-    q: str = Query(..., description="Query string"),
     collection: str = Query(..., description="Collection"),
+    q: str = Query(..., description="Query string"),
```
danielkryska, line 19: This same as above. Docstrings with params or URL to its documentations + examples

backend/app/routes/search.py
```diff
@@ -5,33 +5,39 @@
...
 
 
+# pylint: disable=too-many-arguments
 @internal_api_router.get("/search", name="apis:get-search")
 async def search_get(
-    q: str = Query(..., description="Query string"),
     collection: str = Query(..., description="Collection"),
-    rows: int = Query(10, description="Row count", gte=0, le=100),
-    cursor: str = Query("*", description="Cursor"),
+    q: str = Query(..., description="Query string"),
```
danielkryska, line 17: This same as above. Docstrings with params or URL to its documentations + examples

backend/app/solr/operations.py
```diff
@@ -0,0 +1,31 @@
...
+from httpx import AsyncClient, Response
+
+from app.config import SOLR_URL
+
+
+async def search(
+    client: AsyncClient,
+    collection: str,
+    *,
+    q: str,
```
danielkryska, line 11: This same as above. Docstrings with params or URL to its documentations + examples

backend/app/solr/operations.py
```diff
@@ -0,0 +1,31 @@
...
+    qf: list[str],
+    fq: list[str],
+    sort: list[str],
+    rows: int,
+    cursor: str = "*",
+) -> Response:
+    """Retrieve search results"""
```
danielkryska, line 18: This is the redundant comment I was talking about yesterday. Code is enough self-explanatory.

backend/app/routes/search.py
```diff
@@ -5,33 +5,39 @@
...
+        description="Filter query",
+        example=["journal:Geonomos", 'journal:"Solar Energy"'],
+    ),
     sort: list[str] = Query(
         [], description="Sort order", example=["description asc", "name desc"]
     ),
+    rows: int = Query(10, description="Row count", gte=0, le=100),
+    cursor: str = Query("*", description="Cursor"),
 ):
     """Do a search against the specified collection"""
```
danielkryska, line 30: Would be better to rename the function to `serach_in_collection` or `get_from_collection`  instead of adding a comment? The path will be the same.

backend/app/routes/recommend.py
```diff
@@ -41,40 +39,32 @@ async def recommend_post(
 ):
     """Do a search against the specified collection, pass results to RS"""
```
danielkryska, line 40: Wouldn't be better to call the function `get_collection_recommendations` instead of adding the comment?

danielkryska (approved):

backend/app/routes/search.py
jswk: Here I went for matching the HTTP path section and verb. Docstring can then provide extra info.

PR #33, opened 2022-02-07
fix(auth): ui auth login and logout

danielkryska:


jswk (approved):
Correct the styles though

PR #96, opened 2022-05-24
chore: Refactor services to for search suggestions

### LINKED ISSUE #86: Improve Filters UX
* [x] move active filters above searched results
* [x] improve styles of the active filters

michal-szostak:
* Move search to ELF style repository


ui/libs/search/src/lib/collections/trainings/training.model.ts
```diff
@@ -9,6 +9,7 @@ import {
 } from './training-fields.types';
 
 export interface ITraining {
+  id: string;
```
danielkryska, line 12: Why have you put the id in here, while you're using `Partial<ITraining> & {id: string}`.

ui/libs/layout/src/lib/components/main-header/utils.ts
```diff
@@ -1,21 +1,17 @@
...
-    catchError(() => {
-      window.eosccommon.renderMainHeader(`#${id}`);
-      return of();
-    }),
-    tap(({ username }) =>
-      window.eosccommon.renderMainHeader(`#${id}`, {
-        username,
-      })
+    // required for rerendering to work
+    delay(0),
```
danielkryska, line 12: why are you using the `delay`. I believe it's some kind of async hack to put user update to the end of the stack, but I may be wrong.

ui/libs/search/src/lib/collections/services/adapter.ts
```diff
@@ -1,11 +1,12 @@
 import { IService } from './service.model';
-import { IResult } from '../../result.model';
+import {IResult} from "../../state";
 
 export const serviceToResult = (
-  service: Partial<IService>,
+  service: Partial<IService> & {id: string},
```
danielkryska, line 5: The same as above. You're using `Partial<T> & { id: string }` in adapters, as also id in the models themselves.

ui/libs/search/src/lib/sets/all.set.ts
```diff
@@ -12,4 +15,5 @@ export const allSet: ISet = {
   ],
   urlPath: 'all',
   collections: [servicesCollection, researchProductsCollection],
```
danielkryska, line 17: Will you remove the old field `collections`?

ui/libs/search/src/lib/state/results/results.repository.ts
```diff
@@ -0,0 +1,238 @@
...
+  description: string;
+  type: string;
+  typeUrlPath: string;
+  fieldToFilter: HashMap<string>;
+  collection: string;
+  url: string;
+  fieldsToTags: string[];
+
+  [tagName: string]: string | string[] | any | any[];
+}
```
danielkryska, line 37: please be sure to change the interface to the final version:

```
export interface IResult {
  title: string;
  description: string;
  type: string;
  typeUrlPath: string;
  collection: string;
  url: string;
  tags: ITag[];
}

File: ui/libs/types/src/lib/operators.ts [hunk 6]
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Apply map to observable and map on array using provided function
+ *
+ * @param _func
+ */
+export function mapArray<T, V>(_func: (element: T) => V) {
+  return function(source: Observable<T[]>): Observable<V[]> {
+    return source.pipe(map(array => array.map(_func)));
+  }
+}
```
danielkryska, line 12: Are you sure the operator is simplifying anything? 

`map(array => array.map(_func))` is replaced with `mapArray(_func)`

I believe the added value is too small to use and remember about such a tool in future development.

ui/libs/types/src/lib/types.ts
```diff
@@ -0,0 +1,4 @@
+export type HashMap<T> = { [field: string]: T };
+export interface IHasId {
+  id: string
+}
```
danielkryska, line 4: Why that interfaces/types aren't kept together with all others?

danielkryska:
Are you going to fully replace `collections` with `newCollections`? Or do you want to do it in other PR?

ui/libs/search/src/lib/collections/trainings/training.model.ts
michal-szostak: Because `Partial<ITraining>` can have id or might not have it. The type `Partial<ITraining> & IHasId` WILL have id field which is string.

ui/libs/layout/src/lib/components/main-header/utils.ts
michal-szostak: Exactly - this is equivalent to having instructions wrapped in `setTimeout` they will be executed after current instruction pass. Here it is required because rerendering basically uses another framework so we make sure that everything is ready and only then force the rerender.

ui/libs/pages/search/src/lib/components/filters/filters.component.ts
```diff
@@ -61,11 +58,12 @@ export class FiltersComponent {
   set filters(
     filters:
       | [
-          CollectionSearchMetadata<any>,
+          ICollectionSearchMetadata,
           { [facetName: string]: IFacetResponse }
         ][]
       | null
   ) {
+    console.log(filters)
```
danielkryska, line 66: to be removed, it can be fixed in next PRs

ui/libs/pages/search/src/lib/components/filters/filters.component.ts
```diff
@@ -101,6 +99,7 @@ export class FiltersComponent {
           }
         });
     });
+    console.log(filtersTree)
```
danielkryska, line 102: to be removed, it can be fixed in next PRs

ui/libs/search/src/lib/state/results/results.repository.ts
```diff
@@ -0,0 +1,260 @@
...
+//   fieldsToTags: string[];
+//
+//   [tagName: string]: string | string[] | any | any[];
+// }
+
+// export interface ITag {
+//   type: string;
+//   value: string | string[];
+//   originalField: string;
+// }
```
danielkryska, line 45: if those aren't needed remove them now or in next PRs

danielkryska (approved):

PR #99, opened 2022-05-25
feat: [#84] show up to 3 results per collection

### LINKED ISSUE #84: Show up to 3 results per collection in most relewant
image

michal-szostak:
## WIP

This is work in progress (query to solr itself will be optimized to ask only 3 rows, etc)

## Question

@danielkryska , @agpul - What should happen after clicking the suggested result? Should it work the same as in MP where it goes to the associated URL? What about entries without valid urls? should they be shown?
Additionally - should those suggestions be scoped (All catalogs, Publications, Services...) or should they always be searched within "all catalogs"?

# How does it look like?

image

Closes #84


danielkryska:
> What should happen after clicking the suggested result?
Should redirect to the destination URL as it's with more detailed results on the page.

> Should it work the same as in MP where it goes to the associated URL?
Yes.

> What about entries without valid URLs? should they be shown?
I don't think there will be such a possibility.


> Additionally - should those suggestions be scoped (All catalogs, Publications, Services...) or should they always be searched within "all catalogs"?
Scoped 1:1 like in the Marketplace excluding `All catalogs` scope.

danielkryska (approved):

danielkryska:
When I don't write anything, the box shouldn't occur
image

danielkryska:
Can we add `see all` btn for each collection?

image

danielkryska:
Maybe there should be btn `Show all results` or smth?

danielkryska:
Could you edit styles as such? I don't know when, but it has disappear

```
.container--xxl {
  padding: 25px 50px;
  max-width: 1200px;
  margin: 0 auto;
}

danielkryska:
In case when results will provide to external sources, the titles should be blue.
image

danielkryska (approved):

PR #129, opened 2022-06-29
Add categories

danielkryska:


danielkryska:
TODO
- hook on active id change to build tree
- store active category id in URL Query param
- categories component should have inputs and outputs instead of internal logic and DI of categories service

Issue #140, opened 2022-06-07
Create the Detail Page for Training catalogue entry


michal-szostak:
Metadata that should be taken into consideration during project process of the Training Catalogue Details page: https://wiki.eoscfuture.eu/display/PUBLIC/Training+Catalogue+-+Minimal+Metadata+for+Learning+Resources

PR #151, opened 2022-07-21
[#140] training catalogue

michal-szostak:
https://github.com/cyfronet-fid/eosc-search-service/issues/140

### DIFF
--- a/backend/app/generic/models/presentable.py
+++ b/backend/app/generic/models/presentable.py
@@ -9,19 +9,19 @@ class Presentable(BaseModel):
     id: int | str
     Resource_title_s: str
     Description_s: str
-    Author_ss: list[str]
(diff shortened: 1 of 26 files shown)


Marcelinna:
@agpul Direct download shouldn't be only for Content type pdf ?
image

agpul:
> @agpul Direct download shouldn't be only for Content type pdf ? image

probably should, but unfortunately, we don't have any specifications, so what we can do is guess....

Marcelinna:
> > @agpul Direct download shouldn't be only for Content type pdf ? image
> 
> probably should, but unfortunately, we don't have any specifications, so what we can do is guess....

No "Direct download" redirect to the training page, it should be changed or not for now?

Issue #173, opened 2022-09-01
Clear all filter does not clear the entered phrase


Marcelinna:
Button "Clear all filter" does not clear the entered phase
![image](https://user-images.githubusercontent.com/59872096/187903522-fd4d1066-d3f6-496f-a145-ddd6d8acee37.png)

@agpul Please decide
"Clear all filter" button shouldn't appear when a phrase is typed. It should work like in Marketplace.
![image](https://user-images.githubusercontent.com/59872096/187903885-d7741ac6-d6b0-413c-8fc4-d7c0e7052605.png)

Or "Clear all filter" button should be visible and clear typed phrases and all other filters?

Marcelinna:
@danielkryska it should work like in MP

Issue #164, opened 2022-08-24
Cleaning data


danielkryska:
Pre-requisites
- [ ] what do we expect from data?

TODO
- [ ] remove special chars or records containing it

danielkryska:
Not allowed signs:
- HTML like https://www.toptal.com/designers/htmlarrows/symbols/
- \, /, (, ), [, , ], ', ", |, ...

Allowed
- >, &, ',',

danielkryska:
The issue may be solved by providing better serializers on the UI side.

Marcelinna:
Author name can be 
    'name, surname
    'name surname'
    'surname'
    'name first letter second name. surname' --> Paul E. Wagner

Marcelinna:
![image](https://user-images.githubusercontent.com/59872096/188086067-8cec6d53-702e-4ea4-b2d2-a9bfd880652b.png)
which characters are displayed like this?

PR #186, opened 2022-09-15
Add serializer layer

danielkryska:
Provided changes:
- add validation for filter types & serializers (now filters with special chars like [+\-&|!()"~*?:\\] are escaped) #164 
- Clear all filter does not clear the entered phrase https://github.com/cyfronet-fid/eosc-search-service/issues/173


Marcelinna:
Why link is not added to filter?
image

Marcelinna:
removing the phrase should work just like in MP, so only the button in the search bar is responsible for it
image

michal-szostak (approved):

Marcelinna (approved):

PR #194, opened 2022-09-19
Add colored tags and date to results

danielkryska:


danielkryska:
image

danielkryska:
Changes are provided only for publications.

The open access badge is green, and any other type of access is red.

Date should occur on top of result

Marcelinna (approved):

PR #206, opened 2022-09-27
feat: create all collection based on openaire dump

### LINKED ISSUE #165: Create all_collection based on the new dump from OpenAire
Based on [final data schema create a new dump](https://docs.google.com/document/d/1bav69NTLSS8WFmhO18JbKi568ltOOhA4tUKdS7_i9_g/edit?usp=sharing)

Michal-Kolomanski:
closes #165


transform/all_collection/spark/transform_df/trainings_transform.py
```diff
@@ -1,56 +1,102 @@
...
         .withColumnRenamed("Level_of_expertise_s", "level_of_expertise")
         .withColumnRenamed("License_s", "license")
         .withColumnRenamed("Qualification_s", "qualification")
-        .withColumnRenamed("Resource_Type_s", "document_type")
+        .withColumnRenamed("Resource_Type_s", "resource_type")
         .withColumnRenamed("Resource_title_s", "title")
         .withColumnRenamed("Target_group_s", "target_group")
         .withColumnRenamed("URL_s", "url")
-        .withColumnRenamed("Version_date__created_in__s", "published")
+        .withColumnRenamed("Version_date__created_in__s", "publication_date")
```
danielkryska, line 99: Column rename should be stored in map and then dynamically handled.

Example
```
map = dict(**{|
  "source_field_name": "destination_field_name",
  # ...
  "Auhtor_ss": "author_names"
})

# mapping
for (key of keysOf(map)):
  trainings.wihtColumenRenamed(key, map[key])
```

transform/all_collection/spark/utils/common_df_transformations.py
```diff
@@ -0,0 +1,246 @@
...
+    """Create boolean value whether record is open access or not, based on col_name"""
+    best_access_right = df.select(col_name).collect()
+    open_access_column = []
+
+    for access in best_access_right:
+        if access[col_name] in {
+            "OPEN",
+            "open access",
+            "fully_open_access",
+            "open_access",
```
danielkryska, line 121: In UI I've done it on regex /open.(access)?/gi

transform/all_collection/spark/utils/common_df_transformations.py
```diff
@@ -0,0 +1,246 @@
...
+
+    for instances in instances_list:
+        if instances["instance"]:
+            url_list = []
+            document_type_list = []
+            for instance in instances["instance"]:
+                if instance["url"]:
+                    for url in instance["url"]:
+                        if url and url not in url_list:
+                            url_list.append(url)
```
danielkryska, line 170: I'm a person who likes fast failure and fast return, as also flat code. Did you consider writing it as such?

```
for instances in instances_list:
    if not instances["instance"]:
        continue

    url_list = []
    document_type_list = []
    for instance in instances["instance"]:
        if not instance["url"]:
            continue

        for url in instance["url"]:
            if not url or url in url_list:
                continue

            url_list.append(url)

# or in more map reduce form

non_empty_instances_list = [instance['instance'] for instance in instances_list if instance['instance']] 
for instances in non_empty_instances_list:
    url_list = []
    document_type_list = []
    instances_urls = [instance['url'] for instance in instances if instance['url']]
    for url in instances_urls:
        if not url or url in url_list:
            continue

        url_list.append(url)

File: transform/all_collection/spark/utils/common_df_transformations.py [hunk 4]
```diff
@@ -0,0 +1,246 @@
...
+
+    for countries in countries_list:
+        if countries["country"]:
+            country_list = []
+            for country in countries["country"]:
+                country_list.append(country["code"])
+
+            country_column.append(country_list)
+        else:
+            country_column.append(None)
```
danielkryska, line 196: This could be also simplified

```
for countries in countries_list:
  if countries['country']:
    country_column.append([country['code'] for country in countries['country'])
    continue
  
  country_column,append(None)

danielkryska (approved):
In general, the code is well done, except few clean code things (or simplifications for future you). Those may be treated more as code practices than real issues. 

I know that transformations in data are very dynamic and it isn't worth to write it better, that's why I gave you approve with comments.

transform/all_collection/spark/utils/common_df_transformations.py
Michal-Kolomanski: My approach is more bulletproof. Some records may have `open` in the `best_access_right` and do not be `open access.` I will leave it as it is

transform/all_collection/spark/utils/common_df_transformations.py
Michal-Kolomanski: You are right that this nested code is ugly and could have been avoided. But the suggested code will now work properly,
by design creating dataframes and joining dataframes requires constant number of rows in columns so we can not continue in the loop, instead, we are supposed to append None or an empty array. Order matters. As a result, your suggested code after required changes will look more or less the same as my solution. For now I will keep it as it is. But I will refactor this later on

transform/all_collection/spark/utils/common_df_transformations.py
Michal-Kolomanski: :clap: I have changed `harvest_research_community` also in that way.

danielkryska (approved):

PR #214, opened 2022-09-30
fix: traing detail page

danielkryska:


danielkryska:
image

danielkryska:
image

backend/app/generic/models/presentable.py
```diff
@@ -7,21 +7,20 @@
...
+    language: list[str]
+    keywords: list[str]
+    license: str
+    best_access_right: Optional[str] = "open_access"
+    publication_date: Optional[str] = None
+    resource_type: str
+    content_type: str
+    url: list[str]
+    eosc_provider: str
+    format: list[str]
```
Michal-Kolomanski, line 22: It is risky to assume that most of those fields exist. Many of them can be set to None

Michal-Kolomanski (approved):

PR #222, opened 2022-09-30
feat: stable range filter

danielkryska:


Marcelinna (approved):

ui/apps/ui/src/app/components/filters/filter-range/filter-range.component.ts
```diff
@@ -0,0 +1,149 @@
...
+      .subscribe(async ([start, end]) => {
+        const fqMap = this._customRoute.fqMap();
+        fqMap[this.filter] = [
+          start === this.min ? null : start,
+          end >= this.max ? null : end,
+        ];
+        const filtersConfigs = this._filtersConfigsRepository.get(
+          this._customRoute.collection()
+        ).filters;
+        await this._router.navigate([], {
```
michal-szostak, line 99: having async / await in subscribe is not the best solution this could lead to race condition and parallel execution of two subscription calls

michal-szostak (approved):

goreck888 (approved):

ui/apps/ui/src/app/components/filters/filter-range/filter-range.component.ts
```diff
@@ -0,0 +1,150 @@
...
+          ).filters;
+          return this._router.navigate([], {
+            queryParams: {
+              fq: deserializeAll(fqMap, filtersConfigs),
+            },
+            queryParamsHandling: 'merge',
+          });
+        })
+      )
+      .subscribe();
```
danielkryska, line 106: resolved

michal-szostak (approved):

PR #252, opened 2022-10-14
feat: robust script to transform and send data to solr

### LINKED ISSUE #223: Make robust script to transform and send data to solr

Michal-Kolomanski:
closes #223


transform/all_collection/README.md
```diff
@@ -2,11 +2,18 @@
...
-- `OUTPUT_FORMAT` - Format of the output data files. Default: `JSON`.
\ No newline at end of file
+- `OUTPUT_FORMAT` - Format of the output data files. Default: `JSON`.
+- `SOLR_ADDRESS` - Solr address. Default: `http://127.0.0.1`.
+- `SOLR_PORT` - Solr port. Default: `8983`.
+- `SOLR_DATASET_COLS` - The name of the collection/collections to which datasets will be send. To specify multiple collections, pass them in `""` and separate by a space. Example: `"all dataset"` .
+- `SOLR_PUBLICATION_COLS` - The name of the collection/collections to which publications will be send.
+- `SOLR_SOFTWARE_COLS` - The name of the collection/collections to which software will be send.
+- `SOLR_TRAINING_COLS` - The name of the collection/collections to which trainings will be send.
+- `SOLR_SERVICE_COLS` - The name of the collection/collections to which services will be send.
```
danielkryska, line 19: Do you have any examples? It could help in case of context.

transform/all_collection/README.md
```diff
@@ -2,11 +2,18 @@
...
-- `SOFTWARE_PATH` - path to software data file.
-- `TRAININGS_PATH` - path to trainings data file.
-- `SERVICES_PATH` - path to services data file (optional).
-- `OUTPUT_PATH` - where the output should be saved? Default: `output`
+- `DATASET_DIR_PATH` - path to datasets **directory**.
+- `PUBLICATION_DIR_PATH` - path to publications **directory**.
+- `SOFTWARE_DIR_PATH` - path to software **directory**.
+- `TRAINING_FILE_PATH` - path to trainings **directory**.
+- `SERVICE_FILE_PATH` - path to services **directory**.
+- `OUTPUT_DIR_PATH` - where the output should be saved? Default: `output`.
```
danielkryska, line 10: Paths can have default values.

For example:
DATASET_DIR_PATH: './${OUPUT_DIR_PATH}/datasets'

Won't be easier to have a full path instead DIR_PATH and FILE_PATH. I believe most of the use cases will be to `mv` `OUTPUT` directory content somewhere or load all files inside it.

transform/all_collection/spark/process.py
```diff
@@ -0,0 +1,125 @@
...
+from transform.all_collection.spark.schemas.harvested_props_schemas import (
+    harvested_schemas,
+)
+
+
+def check_trans_consistency(collections: Dict) -> None:
+    """Check whether all data types after transformations will have the schema
+    such that it will be possible to send all data into single solr collection"""
+    dfs = {}
+    for name, _col in collections.items():
```
danielkryska, line 24: Why are you using two different cases while unpacking? 

`name`

`_col`, with an underscore

transform/all_collection/README.md
Michal-Kolomanski: I can set some defaults but in my opinion, they won't be so much useful. Maybe instead of them, `env template` will be better? A file with more or less my current setup:
```
DATASET_DIR_PATH=input_data/dataset/
PUBLICATION_DIR_PATH=input_data/publication/
SOFTWARE_DIR_PATH=input_data/software/
TRAINING_FILE_PATH=input_data/trainings//
SERVICE_FILE_PATH=input_data/services/
OUTPUT_DIR_PATH=output_data
INPUT_FORMAT=json
OUTPUT_FORMAT=json
SOLR_ADDRESS=http://127.0.0.1
SOLR_PORT=8983
SOLR_DATASET_COLS="all_collection_26-09-2022 dataset_26-09-2022"
SOLR_PUBLICATION_COLS="all_collection_26-09-2022 publication_26-09-2022"
SOLR_SOFTWARE_COLS="all_collection_26-09-2022 software_26-09-2022"
SOLR_TRAINING_COLS="all_collection_26-09-2022 training_26-09-2022"
SOLR_SERVICE_COLS="all_collection_26-09-2022 service_26-09-2022"

```
Always there is an option to add defaults and env.template. What is your opinon?

Those **input** directories like`DATASET_DIR_PATH`, `PUBLICATION_DIR_PATH`, etc. can not be placed inside `OUPUT_DIR_PATH` - this is the **output** directory which is cleared entirely in every iteration of the program. 
But what your comment remind me of is changing names for:
`TRAINING_FILE_PATH` -> `TRAINING_DIR_PATH`
`SERVICE_FILE_PATH` -> `SERVICE_DIR_PATH`
In that case I will just remove "DIR" part from all variables just to make it shorter

Michal-Kolomanski: In the next commit I will add:
1) defaults for input data
2) env template
3) make env path variables shorter

transform/all_collection/spark/process.py
Michal-Kolomanski: `_col`  - I used here an underscore because col is a method in some packages like `pyspark,` `pandas` etc. So it potentially may shadow variable from outer scope.

I have changed those variables to:
`name` -> `collection_name` (col_name is in outer scope so it should not be used here)
`_col` -> `collection`

transform/all_collection/README.md
Michal-Kolomanski: Ok I have added examples to each

danielkryska (approved):
Generally good-looking PR. Although there are a few clean code issues:

- literal values instead of const (`2`, `1`)
- too large methods
- mixed level of abstraction

There are also few things related with your latest PR:
- missing inline `for`
- missing inline `if else`
- missing inline default values `x or 'my default string'`

Try to use [`map`, `reduce`, and `filter`](https://book.pythontips.com/en/latest/map_filter.html). It can simplify significantly your code and also optimize it (because those functions return generators)! :)

transform/all_collection/spark/utils/utils.py
```diff
@@ -47,3 +46,23 @@ def add_columns(df: DataFrame, cols: Tuple) -> DataFrame:
...
+    assert all(
+        (columns == df_columns[0] for columns in df_columns)
+    ), "Dataframes after transformation don't have the same columns name"
+
+    dfs_cols_schema = [[column.dataType for column in df.schema.fields] for df in dfs]
+
+    for column in zip(*dfs_cols_schema):
+        column_schemas = set((str(_col.typeName()) for _col in column))
+        # Columns in all dataframes should have the same type or void + other type
+        if len(column_schemas) == 2:
```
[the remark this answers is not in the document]
danielkryska: I'm not sure if you shouldn't use `is` instead of `==`, but I'm not the python master.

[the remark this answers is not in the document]
Michal-Kolomanski: Python caches integer objects in the range -5..256 as singleton instances for performance reasons - so it will work in that case, BUT it is not a good practice to relay on that. In that case, we should use the equality operator `==`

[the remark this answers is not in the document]
Michal-Kolomanski: Hmmm I do not know if I should create some variables for those 2 and 1. The case here is a little bit confusing but in general:
Here we want to check 
- if all columns in each dataframe have the same name as the corresponding column in other dataframes
- if all columns in each dataframe have the same type (`schema length == 1`)(OR if there is a case in which one column in df1 is void type, and the same column in df2 is something different than void - it's also good - `length of schemas == 2`) as the corresponding column in other dataframes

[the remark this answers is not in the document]
Michal-Kolomanski: I have switched statements a little bit. Do you have some tips to make it cleaner? One line if else statement is too long I guess.
image

PR #264, opened 2022-10-19
chore: update readme release section

danielkryska:


Michal-Kolomanski (approved):
Thanks for providing the links. The only thing that may be missing is the part about adjusting the changelog, please make a short note on which stage we may adjust that. I believe that @wziajka also should look at this, maybe adding him to reviewers is worth considering?

wziajka (approved):

Michal-Kolomanski (approved):

PR #269, opened 2022-10-21
feat: add doi

### LINKED ISSUE #250: Improve the DOI in the Search Service
Check one more time the dump from the OpenAire and extract the DOI according to the Miriam clues.

Michal-Kolomanski:
closes #250

When **reviewing** please look only at the first commit. 
The second one just adds the newest Solr schema (field `doi` was added as a type of strings - in Python this field is a list of strings)

**Required actions:** 
- First of all, data have to be sent to the production solr instance - currently, it is NOT, I will do it probably this afternoon after finishing my PR with the provider also. It takes more or less 10 hours to send the data :( 

**So on Monday:** @danielkryska  @michal-szostak 
- Solr schema has changed so locally unfortunately you also have to change your schema,
- Frontend should display DOI using the new field `doi` not `url` as it is today

**Tested locally (works):**
image


transform/all_collection/spark/transformations/commons.py
```diff
@@ -242,6 +241,22 @@ def harvest_research_community(df: DataFrame, harvested_properties: Dict) -> Non
...
 
+def harvest_doi(df: DataFrame, harvested_properties: Dict) -> None:
+    """Harvest DOI from OAG resources"""
+    pids_list = df.select("pid").collect()
+    doi_column = []
+
+    for pids in pids_list:
+        if pids["pid"]:
+            doi_column.append(
+                [pid["value"] for pid in pids["pid"] if pid["scheme"] == DOI]
```
danielkryska, line 252: Good use of `inline for`. You may also put the `[pid["value"] for pid in pids["pid"] if pid["scheme"] == DOI]` to separate variable to improve readability. It's important to take care of that part since we - software developers - spend 10 times more time reading than writing the code. 

Example solution for better readability:

```python
doi_urls = [pid["value"] for pid in pids["pid"] if pid["scheme"] == DOI]
doi_column.append(doi_urls)

Comment (danielkryska) [reply transform/all_collection/spark/transformations/commons.py:252]: Are you sure that the `value`, and `scheme` keys are always available in `pid`? Won't the code throw `KeyError`?

Comment (danielkryska) [reply transform/all_collection/spark/transformations/commons.py:252]: You could simplify the code significantly using default values. It's very good practice in the case of transforming arrays.

Source of solution:
https://stackoverflow.com/questions/13710631/is-there-shorthand-for-returning-a-default-value-if-none-in-python

Final example: 

```python
for pids in pids_list:
    pids_values = pids["pid"] or []
    doi_urls = [pid["value"] for pid in pids_values if pid["scheme"] == DOI]
    doi_column.append(doi_urls)
```

When `pids['pid']` is None then `pids_values` will equal `[]`
When `pids_values` is `[]`, then `doi_urls` are equal to `[]`

danielkryska (approved):
Every time we speak about data you always say: `expect anything in the data`. :) I think that maybe in the case of missing `scheme` or `value` we should also be secured. 

Great work with the `inline for`. I see an enormous improvement in this case. Keep it that way.

When you transform arrays try to use the default value (`[]`) when the variable you want to transform is `None` (more details I wrote in the code comments).

I'll give you approval since I believe that changes are fast and small to be provided.

transform/all_collection/spark/transformations/commons.py
Michal-Kolomanski: > Are you sure that the `value`, and `scheme` keys are always available in `pid`? Won't the code throw `KeyError`?

Generally, yes. If a `pid` exists the `value` and `schema` keys should always exist too. If that assumption wasn't true, that will mean that OpenAire data is invalid in that case. But I agree that we can be prepared for that.

Michal-Kolomanski: > You could simplify the code significantly using default values. It's very good practice in the case of transforming arrays.
> 
> Source of solution: https://stackoverflow.com/questions/13710631/is-there-shorthand-for-returning-a-default-value-if-none-in-python
> 
> Final example:
> 
> ```python
> for pids in pids_list:
>     pids_values = pids["pid"] or []
>     doi_urls = [pid["value"] for pid in pids_values if pid["scheme"] == DOI]
>     doi_column.append(doi_urls)
> ```
> 
> When `pids['pid']` is None then `pids_values` will equal `[]` When `pids_values` is `[]`, then `doi_urls` are equal to `[]`

Yeah, great idea to use `or` for default values. 
I have applied your suggestion! :clap:

PR #274, opened 2022-10-25
fix: add missing tags

### LINKED ISSUE #235: Lack of Scientific domain, Organisation for Services in All Catalogs
All Catalogs
image

Services
image

danielkryska:
closes #235

### DIFF
--- a/ui/apps/ui/src/app/collections/data/all/adapter.data.ts
+++ b/ui/apps/ui/src/app/collections/data/all/adapter.data.ts
@@ -64,8 +64,18 @@ export const allCollectionsAdapter: IAdapter = {
       },
       {
         label: 'DOI',
-        value: data?.url || [],
-        filter: 'url',
+        value: data?.doi || [],
+        filter: 'doi',
+      },
+      {
+        label: 'Scientific domain',
+        value: data?.scientific_domains || [],
+        filter: 'scientific_domains',
+      },
+      {
+        label: 'Organisation',
+        value: data?.resource_organisation || '',
+        filter: 'resource_organisation',
       },
     ],
     type: data?.type || '',
--- a/ui/apps/ui/src/app/collections/data/all/filters.data.ts
+++ b/ui/apps/ui/src/app/collections/data/all/filters.data.ts
@@ -35,9 +35,21 @@ export const allCollectionsFilters: IFiltersConfig = {
       type: 'tag',
     },
     {
-      id: 'url',
-      filter: 'url',
+      id: 'doi',
+      filter: 'doi',
       label: 'DOI',
+      type: 'tag',
+    },
+    {
+      id: 'scientific_domains',
+      filter: 'scientific_domains',
+      label: 'Scientific Domains',
+      type: 'tag',
+    },
+    {
+      id: 'resource_organisation',
+      filter: 'resource_organisation',
+      label: 'Resource organisation',
       type: 'tag',
     },
   ],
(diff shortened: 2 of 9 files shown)


danielkryska:
image

danielkryska:
image

danielkryska:
image

danielkryska:
image

danielkryska:
image

ui/apps/ui/src/app/collections/data/all/adapter.data.ts
```diff
@@ -64,8 +64,18 @@ export const allCollectionsAdapter: IAdapter = {
       },
       {
         label: 'DOI',
-        value: data?.url || [],
-        filter: 'url',
+        value: data?.doi || [],
```
Michal-Kolomanski, line 67: @agpul `DOI` occurs only for Openaire-type resources. What do we want to do with trainings and services?
Do we want to present nothing or a `URL`?

ui/apps/ui/src/app/collections/data/all/adapter.data.ts
```diff
@@ -64,8 +64,18 @@ export const allCollectionsAdapter: IAdapter = {
...
+        filter: 'doi',
+      },
+      {
+        label: 'Scientific domain',
```
Michal-Kolomanski, line 71: @agpul  `Scientific domain` is only for services! For OAG-type resources, there is `fos` (field of science) which looks similar at least to me. Please look at this filter [Explore](https://explore.openaire.eu/search/find?resultbestaccessright=%22Open%2520Access%22&active=result). Maybe we can present `Scientific domain` for services and `fos` for OAG? For trainings most likely there is nothing like this

Michal-Kolomanski (commented):
The code looks good but I have a few business concerns. Before merging it also be safer, to change aliases to the latest data to enable testing on the PR instance new feature (as I was writing)

ui/apps/ui/src/app/collections/data/all/adapter.data.ts
agpul: nothing:)

ui/apps/ui/src/app/collections/data/all/adapter.data.ts
agpul: in the future, this field will be also valid for training, Scientific domain = FOS from my perspective, and in all catalogs it should be visible as one filter

ui/apps/ui/src/app/collections/data/all/adapter.data.ts
Michal-Kolomanski: ok so it's good as it is

ui/apps/ui/src/app/collections/data/all/adapter.data.ts
Michal-Kolomanski: ok, so this should be done @danielkryska

agpul: #279 - read and comment on what you think. I've created another ticket to resolve it as a new issue.

Michal-Kolomanski: ping @danielkryska

agpul (approved):

Michal-Kolomanski (commented):
@agpul I am also wondering if we should color our DOIs in some way. Currently, it is black as everything. For me, it is not obvious that is an interactive link to the resource. Look how OpenAire is presenting DOIs - as blue links
image

Michal-Kolomanski (approved):

Marcelinna (approved):

PR #309, opened 2022-11-08
feat: add eosc favicon

michal-szostak:
* add eosc favicon
* change research categories to research steps


Michal-Kolomanski (approved):
Before merging, let's fix the failing test

PR #313, opened 2022-11-09
chore: add configurable prefix to collections

### LINKED ISSUE #307: Collections names should be configurable with env variables
**Currently**
All collection names are hardcoded in `search-metadata` files.

**What do we want?**
We want to have a possibility to change them on demand with env variables.

danielkryska:
- add prefixes to collections, by default `NG_COLLECTIONS_PREFIX=''`

Building pipe updates for @wziajka 
- [ ] add the new environment variable to the .env file in the format: `NG_COLLECTIONS_PREFIX=<collection_prefix>`
Example: `NG_COLLECTIONS_PREFIX=prod_` for `production`
- [ ] before building the UI package, copy the `.env` file to the `<root>/ui` catalog

closes #307

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,13 @@ Run `docker-compose up --build`.
 `npm i --force`
 
 ### Run
-**IMOPRTANT!!! To change collections prefix copy `<root>/.env` file to `<root>/ui` with proper env variable.**
+**IMOPRTANT!!! To change collections prefix copy `<root>/.env` file to `<root>/ui` with `NG_COLLECTIONS_PREFIX` env variable.**
 
 `npm start`
 
 ### Build
 Build artifacts can be found in `ui/dist/apps/ui`.
-**IMOPRTANT!!! To change collections prefix copy `<root>/.env` file to `<root>/ui` with proper env variable.**
+**IMOPRTANT!!! To change collections prefix copy `<root>/.env` file to `<root>/ui` with `NG_COLLECTIONS_PREFIX` env variable.**
 
 `npm build`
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -11,10 +11,13 @@ Run `docker-compose up --build`.
 `npm i --force`
 
 ### Run
+**IMOPRTANT!!! To change collections prefix copy `<root>/.env` file to `<root>/ui` with proper env variable.**
```
Michal-Kolomanski, line 14: Be here more precise: 
... with proper `NG_COLLECTIONS_PREFIX` value. 
Did you mean that here?
You can also give an example of such value

README.md
```diff
@@ -11,10 +11,13 @@ Run `docker-compose up --build`.
...
+
 `npm start`
 
 ### Build
 Build artifacts can be found in `ui/dist/apps/ui`.
+**IMOPRTANT!!! To change collections prefix copy `<root>/.env` file to `<root>/ui` with proper env variable.**
```
Michal-Kolomanski, line 20: same as above

Michal-Kolomanski (approved):
So the assumption is that we will establish a convention by which production and dev instances will **always** use certain aliases. If some for example PR will require other aliases, @wziajka will be able to adjust that on the fly. 

Before merging:
- let's agree on prefixes that will be used from now, and where @wziajka @Michal-Kolomanski

README.md
danielkryska: done

README.md
danielkryska: done

wziajka (approved):

Issue #272, opened 2022-10-24
Improve the behavior of the search bar in the Search Service


agpul:
The scope of this ticket is to make a full check of the search bar behavior and improve it. One issue that has been found during the live demo:
1.  Go to: https://search.eosc-portal.eu/
2. Type for eg: "amnesya" - see no result
3. Type: amnesia - see results
 As you can see if the user makes a mistake in typing the word there are no results.

danielkryska:
https://solr.apache.org/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-FuzzySearches
https://solr.apache.org/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-ProximitySearches
https://solr.apache.org/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-ConstantScorewith_
https://solr.apache.org/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-TheBooleanOperator_

danielkryska:
wild cards: + (single char), * (multiple chars)
long terms with white chars have been wrapped into "

biometricsmw:
It seems that ~ is a good solution to show plenty of results. However the main problem here is that it is fuzzy search - so when we will try to observe "int" word (as the beginning of word "Introduction") we will also gain some results were "in" is observable. It is related to the methodology used for measuring the distance between the term and other words. When it comes to the possible solution to this problem is increasing the validity of the term with ^ - e.g., int^5 will tell solr that term int is important and should be searched as the first one. However, it seems that only combination of these two solutions can work appropriately. It is especially observable in the case of multiple terms - e.g., "int t" will show 0 results, "int t~" will show almost all results from the database, "int^5 t~" will reduce the number of the found records. However, we should also consider the importance of the terms when there are multiple of them. I would recommend to use the length of the term as "importance level". So our latest search will be "int^3 t^1 ~". 

To sum up:
- "~" is needed to run fuzzy search
- "^" is needed to assign importance level - to reduce the number of found records (importance level can be assigned on the basis of word length)

However, it is also interesting how "+" (required) will work. For now, it is removed from query string.

biometricsmw:
Also ~ works well when it comes to typo in the words:
- Write q=persimtent -> no results
- Write q=persimtent~ -> the results are shown as well as they include those that are required by the user

danielkryska:
Use cases for searching:

Software:
[...]

PR #315, opened 2022-11-09
[#272] improve the behavior of the search bar

danielkryska:
https://github.com/cyfronet-fid/eosc-search-service/issues/272


danielkryska:
Except for using single `~` char, you proposed more sophisticated solutions after your research what happened to them?

danielkryska:
In case of checklist before [`review`](https://github.com/cyfronet-fid/eosc-search-service/pull/315#issuecomment-1308823036). You changed only 1 place in the code with `metadata.queryMutator`. That's caused critical building errors in `aot` and `jot`. 

But you'll know it if you tried your solution first.

image

danielkryska:
```bash
Error: apps/ui/src/app/collections/data/datasets/search-metadata.data.ts:23:3 - error TS2322: Type '{ id: string; facets: { document_type: { field: string; type: "terms"; limit: number; }; best_access_right: { field: string; type: "terms"; limit: number; }; language: { field: string; type: "terms"; limit: number; }; ... 6 more ...; research_community: { ...; }; }; queryMutator: (q: string) => string; params: { ......' is not assignable to type 'ICollectionSearchMetadata'.
  Object literal may only specify known properties, and 'queryMutator' does not exist in type 'ICollectionSearchMetadata'.

23   queryMutator: (q: string) => q,
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Error: apps/ui/src/app/components/filters/filter-multiselect/utils.ts:16:15 - error TS2339: Property 'queryMutator' does not exist on type 'ICollectionSearchMetadata'.

16   q: metadata.queryMutator(q),
                 ~~~~~~~~~~~~

biometricsmw:
Hi @danielkryska, thank you very much for your valuable comments and all remarks. When it comes to the more sophisticated methodology of search, once again checked the results of it and it seems that my first impression was wrong. Usage of ^ does not change anything in our case (this impression was connected with the fact that the list of found records changes slightly - but only in order, no new records are added neither any is removed). I also once again read the documentation of the SOLR search and it seems that in our case (because the issue was related to mistake in the word - e.g., "algorihtm" instead of "algorithm") ~ will do the job.

Yesterday, I observed that not all files were added by me to the commit - I am extremely sorry for that - it was the main point of the issue. Before our daily, I will once again check the files and then prepare another commit.

danielkryska:
@biometricsmw, thank you for your additional comment about in-depth solution research. 

Before the next `review` of the Pull Request, please:
- go through all checks listed above in the comments
- write in comments how `~` affect multiple words? Have you checked it?

In case of further review, it will be provided by @michal-szostak. In case of any concerns, ask him questions.

danielkryska:
In the case of multiple words. I have tested `hw to` in `training` to find records starting with `How to ...`.

`hw to~`, which you proposed gave me 0 results
`"hw to"~`, gave me the result I wanted

biometricsmw:
Hi @danielkryska , once again thanks for your comments! When it comes to the example you provided - I also perform it on my latest implementation - the one without "", and received plenty of results. Please take a look at the screenshot attached.
Screenshot 2022-11-10 091236

eldenkat:
After entering the first letters of the name, four results with no relation pop up, only then the right person

biometricsmw:
It seems that (from the code analysis) - the highest priority is for the title, then the authors are considered and finally the description. So it can be a reason why such strange results are also given. Right now, I prepared the same procedure as in the case of suggestions. It is more precise than previously proposed solution. However, I am waiting for your comments.

danielkryska:
Maybe you should update the field's weights in `search-metadata.data.ts` > in the field `qf`.

danielkryska:
You could also add an additional query parameter to a search request `q.op=AND`.

PR #327, opened 2022-11-10
fix: minor bug fixes

danielkryska:
* restore service url in all collection
* restore data source url in all collection
* disable site reloading on data source btn click when in data source collection tab
* fix redirect of data source from all collections


danielkryska:
image

image

image

michal-szostak (approved):

PR #345, opened 2022-11-15
feat: add possibility to change between aai endpoints

danielkryska:
@wziajka set `OIDC_AAI_NEW_API: true` for instances using `AAI demo` or `AAI dev`, in other cases you don't need to change anything.


michal-szostak (approved):

Marcelinna:
@danielkryska it is ok, when I try to log in ?
image

agpul (approved):

Marcelinna (approved):

PR #360, opened 2022-11-18
feat: improve suggestions of search bar

biometricsmw:
In this commit, only search suggestions are improved. In separate commit (to separate branch), I will also provide the needful changes in the case of search itself.

### DIFF
--- a/ui/apps/ui/src/app/components/search-input/search-input.service.ts
+++ b/ui/apps/ui/src/app/components/search-input/search-input.service.ts
@@ -43,7 +43,7 @@ export class SearchInputService {
   _suggestedResultsBy$(q: string, collections: ICollectionSearchMetadata[]) {
     const nq = q.split(' ');
     let new_query = '';
-    nq.forEach(function(value) {
+    nq.forEach(function (value) {
       value = value + '~1';
       new_query = new_query + value + ' ';
     });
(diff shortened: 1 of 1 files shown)


eldenkat:
- [ ] In the results, in addition to the correct ones, items that don't contain the searched phrase at all, are also displayed.

biometricsmw:
One little question, what is the importance of the fields for the search? Because, right now, I changed the implementation of the fuzzy search (only one difference per word is allowed), however, still it shows some strange records. Do we only consider title or other fields (as authors or description) are also considered?

biometricsmw:
@eldenkat could you please take a look at the change? Does it improve the way it the searchbar work?

biometricsmw:
For example I wrote word "perspective" and got the result that (in title) does not contain this phrase - however it is available in the description. Please, take a look at the attached screenshot.
Screenshot 2022-11-21 115545

Marcelinna:
image
word: minh doan does not appear in result prompts

eldenkat:
> @eldenkat could you please take a look at the change? Does it improve the way it the searchbar work?

Yup, ABC and some other phrases are ok. But still smth is wrong, I guess.

biometricsmw:
Thanks @Marcelinna  and @eldenkat for your comments. I assume that the problem lies in the part of the code (qf and numerical values after ^ sign) attached to this comment. However, I need some additional time to check how to fix it (so that it will not destroy the whole search :D ). I will try to experiment with different values for each term (and then we will see whether it is better or not :) ).
Screenshot 2022-11-21 at 16 31 14

biometricsmw:
However, it seems that e.g., abcd-ns was found (for term "resear") because it contains "Research" word in the information :) 
Screenshot 2022-11-21 at 16 36 46

biometricsmw:
@Marcelinna, @michal-szostak  prosiłbym o review

biometricsmw:
This PR will be merged with PR from 272

Issue #324, opened 2022-11-10
Improve the autocomplete in Search Service


agpul:
Improve the behavior of the autocomplete, for eg.
- [ ] when I type "amnes" I see nothing but I would like to see "amnesia"
- [ ] I would like to see also "amnesia" when I write this word with misspellings for eg "amnesya"
- [x] when I am on the all catalog tab and I write "amnes" I can see two two clues from two tabs called "PUBLICATION"
![image](https://user-images.githubusercontent.com/43000481/201077187-6d001cf9-948b-4bf2-b5fc-c3c9c6aca44a.png)
the first one probably refers to all catalog and I would like to ask you to REMOVE it

biometricsmw:
Hi @agpul, could you please check whether this duplication is still an issue? I checked with "amnes" word with the latest version of the app... and cannot observe the problem (please see attached screenshot). 
![Screenshot 2022-11-17 123020](https://user-images.githubusercontent.com/11648507/202435246-bc16ac7b-5143-4bfd-ae08-8f12f4c1be38.png)

agpul:
@biometricsmw you are right, so this bug is resolved

biometricsmw:
In this part, yup, exactly :) But two others are still in progress :)

biometricsmw:
Hi @agpul, as an attachment to this comment I placed an example how the search works for now. As an example, I placed "Cases focarditis" instead of "Case of carditis". Could you please tell me whether such results are acceptable or should I try something to narrow them down? Thanks in advance for your response and opinion! 
![Screenshot 2022-11-18 135344](https://user-images.githubusercontent.com/11648507/202710093-d5c9e357-27f9-403e-8fe5-710e302825fb.png)

agpul:
@biometricsmw Looks ok, but I need to test it on the PR to be 100% sure :) so please make a PR and ma and @Marcelinna will test it.

biometricsmw:
Sure! I will just check one more strategy and then prepare a PR :)

PR #369, opened 2022-11-24
Improvement of search & search suggestions

biometricsmw:
This is the PR that resolves two issues raised in our backlog:

- Improvement of the search bar - to search more results #272 
- Improvement of search suggestions - directly the same way of working as in the case of search bar #324

### DIFF
--- a/ui/apps/ui/src/app/components/filters/filter-multiselect/utils.ts
+++ b/ui/apps/ui/src/app/components/filters/filter-multiselect/utils.ts
@@ -1,3 +1,4 @@
+import { queryChanger } from '@collections/filters-serializers/utils';
 import {
   ICollectionSearchMetadata,
   IFacetParam,
@@ -13,7 +14,7 @@ export const toSearchMetadata = (
   fq: string[],
   metadata: ICollectionSearchMetadata
 ) => ({
-  q: q,
+  q: queryChanger(q),
   fq,
   cursor: '*',
   rows: 0,
--- a/ui/apps/ui/src/app/pages/search-page/search-page.component.ts
+++ b/ui/apps/ui/src/app/pages/search-page/search-page.component.ts
@@ -14,6 +14,7 @@ import {
 import { MAX_COLLECTION_RESULTS } from '@components/results-with-pagination/pagination.repository';
 import { ActivatedRoute, Router } from '@angular/router';
 import { combineLatest } from 'rxjs';
+import { queryChanger } from '@collections/filters-serializers/utils';
 
 @UntilDestroy()
 @Component({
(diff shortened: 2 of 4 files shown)


biometricsmw:
The main issue that was (at least slightly fixed) in this PR is related to search. Right now, it was improved so that slight mistake (e.g., amnesya instead of amnesia) will not have an impact on the search result - the records will be shown.

biometricsmw:
@michal-szostak , @Marcelinna  please review :)

ui/apps/ui/src/app/components/filters/filter-multiselect/utils.ts
```diff
@@ -43,3 +43,13 @@ export const search = (query: string | null, entities: IFilterNode[]) => {
...
+
+export const query_changer = (q: string) => {
+  const nq = q.split(' ');
+  let new_query = '';
+  nq.forEach(function (value) {
+    value = value + '~1';
+    new_query = new_query + value + ' ';
+  });
+  return new_query;
+};
```
danielkryska, line 55: 1. We are using `camelCase` in the frontend, so `query_changer` should be `queryChanger`
2. Maybe instead of using `nq`, you should use `mutatedQuery`, `changedQuery`
3. You're code can be simplified significantly like

```javascript
export const queryChanger = (q: string): string => q.split(' ')
  .map((word) => `${word}~1`)
  .join(' ');
```

4. As you're using the method in multiple places it should be exported from some `utils.ts` file. You can put the code into `ui/apps/ui/src/app/collections/filters-serializers/utils.ts`.

ui/apps/ui/src/app/pages/search-page/search-page.component.ts
```diff
@@ -138,4 +138,14 @@ export class SearchPageComponent implements OnInit {
...
+
+  query_changer = (q: string) => {
+    const nq = q.split(' ');
+    let new_query = '';
+    nq.forEach(function (value) {
+      value = value + '~1';
+      new_query = new_query + value + ' ';
+    });
+    return new_query;
+  };
```
danielkryska, line 150: The same as above.

danielkryska (changes requested):

danielkryska:
Please add changes from the PR https://github.com/cyfronet-fid/eosc-search-service/pull/360 to this one, since both of them are using the same code.

danielkryska:
And after merging, please use the same changes as in my request for the new lines.

danielkryska:
Summary
- [ ] update the code to 
```javascript
export const queryChanger = (q: string): string => q.split(' ')
  .map((word) => `${word}~1`)
  .join(' ');
```
- [ ] move it to `ui/apps/ui/src/app/collections/filters-serializers/utils.ts`
- [ ] rebase all your commits
- [ ] your 1 commit should begin from `feat:` not `chore:`, since `chore:` is reserved for changes that are influencing `build, ci, cd, readme, ...`
- [ ] update properly Pull Request title with what you are actually solving
- [ ] add related ticket to Pull Request `Development` section

danielkryska (approved):

Marcelinna (approved):

PR #391, opened 2022-12-09
refactor: transform

### LINKED ISSUE #298: Refactor data transform part of the repository

Michal-Kolomanski:
closes #298


transform/transformations/common.py
```diff
@@ -22,87 +24,35 @@
...
-# Mappings - values are mapped to the keys
-# Access rights mapping
-_OPEN_ACCESS = "Open access"
-RESTRICTED = "Restricted"
-ORDER_REQUIRED = "Ordered required"
-LOGIN_REQUIRED = "Login required"
-LOGIN_REQUIRED_ON = (
-    "Login required on EOSC Pillar, open access on the original resource page"
+from transform.schemas.mappings import (
+    _OPEN_ACCESS,
```
michal-szostak, line 40: If we import `_OPEN_ACCESS` from other modules it should not be prefixed with `_`

transform/transformations/common.py
```diff
@@ -197,9 +147,12 @@ def map_best_access_right(
     return df.drop(BEST_ACCESS_RIGHT)
 
 
-def create_open_access(best_access_right: List, harvested_properties: Dict) -> None:
+def create_open_access(harvested_properties: Dict) -> None:
```
michal-szostak, line 150: with python 3.10 we can just use `dict` as type

transform/transformers/base/base.py
```diff
@@ -0,0 +1,93 @@
...
+from transform.transformations.common import add_tg_fields
+
+
+class BaseTransformer(ABC):
+    """Base transformer class"""
+
+    def __init__(
+        self,
+        desired_type: str,
+        cols_to_add: Tuple,
```
michal-szostak, line 18: As with `dict` in python 3.10 we can just use `tuple` as type

transform/transformers/base/base.py
```diff
@@ -0,0 +1,93 @@
...
+        cols_to_drop: Tuple,
+        cols_to_rename: Dict,
```
michal-szostak, line 20: If we're typing dict it's good to type what type is the key and value `dict[str, Any]`.

transform/transformers/base/marketplace.py
```diff
@@ -0,0 +1,126 @@
...
+    def apply_simple_trans(self, df: DataFrame) -> DataFrame:
+        """Apply simple transformations.
+        Simple in a way that there is a possibility to manipulate the main dataframe
+        without a need to create another dataframe and merging"""
+        df = df.withColumn(TYPE, lit(self.type))
+        df = self.rename_cols(df)
+        df = self.simplify_urls(df)
+        # Increase the range of data sources IDs -> to avoid a conflict with service IDs
+        if self.type == DATA_SOURCE_TYPE:
+            df = df.withColumn(ID, (col(ID) + 10_000_000))
```
michal-szostak, line 46: This should be extracted to constant - otherwise it becomes magic number.

Maybe even it might be wrapped into separate function `map_id(df) -> str`, or something.

transform/transformers/base/marketplace.py
```diff
@@ -0,0 +1,126 @@
...
+                "link_research_product_metadata_license_urls",
+                "research_product_licensing_urls",
+            )
+
+        for urls in url_cols_to_simplify:
+            df = df.withColumn(urls, col(urls)[URL])
+        return df
+
+    @property
+    def harvested_schema(self):
```
michal-szostak, line 81: Return type not present

michal-szostak (approved):
There is a couple of small issues, but overall it's good.

PR #405, opened 2022-12-22
fix: change type of user action from dict to string

michal-szostak:


backend/app/dependencies/user_actions.py
```diff
@@ -57,19 +57,22 @@ def send(
...
-                    session.aai_state,
-                    url,
-                    session.session_uuid,
-                    resource_id,
-                    resource_type,
-                    page_id,
-                    recommendation,
-                )
-            ),
+            json.dumps(f"{message}"),
```
Michal-Kolomanski, line 75: Is an f-string here necessary?

Michal-Kolomanski (approved):
I compared these UAs to those from Marketplace, and the format & schema seems to be the same.

PR #419, opened 2023-01-17
fix: recommender system

### LINKED ISSUE #417: Fix the recommender
@agpul 
- [x] recommendations response (from our facade) is extended with the boolean flag `isRandom`
- [x] if any of the ids from the recommender can't be found, return all well-suited objects for user needs
- [x] fix recommendations facade based on changes in the external recommender endpoint request body/response
- [x] remove turn on the random flag from env
- [x] add a message parallel to the random flag to inform about the error from the recommender side, when random generated

@danielkryska 
- [x] add `services` recommendations request
- [x] add `data-sources` recommendations request

danielkryska:
closes #417

### DIFF
--- a/backend/app/routes/web/recommendation.py
+++ b/backend/app/routes/web/recommendation.py
@@ -1,4 +1,4 @@
-# pylint: disable=missing-function-docstring
+# pylint: disable=missing-function-docstring,too-many-return-statements
 
 """ Presentable items UI endpoint """
 import datetime
@@ -11,14 +11,10 @@ from typing import Literal
 import httpx
 from async_lru import alru_cache
 from fastapi import APIRouter, HTTPException, Request
-from httpx import AsyncClient
+from httpx import AsyncClient, ReadTimeout
 from starlette.status import HTTP_200_OK
 
-from app.config import (
-    RECOMMENDER_ENDPOINT,
-    SHOW_FIXED_RECOMMENDATIONS,
-    SHOW_RECOMMENDATIONS,
-)
+from app.config import RECOMMENDER_ENDPOINT, SHOW_RECOMMENDATIONS
 from app.generic.models.bad_request import BadRequest
 from app.schemas.session_data import SessionData
 from app.solr.operations import get, search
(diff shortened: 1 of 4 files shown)


tomaszandel (approved):

backend/app/routes/web/recommendation.py
```diff
@@ -76,47 +72,45 @@ def __str__(self):
...
-            "page_id": page_id,
-            "panel_id": panels,
+            "page_id": "/search/" + panel_id,
+            "panel_id": _get_panel(panel_id),
             "candidates": [],
             "search_data": {},
         }
 
+        if session:
+            request_body["aai_id"] = session.aai_id
```
michal-szostak, line 112: In user actions as `aai_uid` `session.aai_state` is being sent. Is this correct? @Michal-Kolomanski ?

Michal-Kolomanski: Yes, it is. But what is the exact difference between `session.aai_state `and `session.aai_id`?

michal-szostak: `state` zwracany jest przez query param z checkina, `aai_id` jest brany z `aai_id = aai_response["userinfo"]["sub"]`

```
async def auth_checkin(code: str, state: str):
    if not state:
        return RedirectResponse(status_code=400, url=UI_BASE_URL)

    try:
        aai_response = rp_handler.finalize(OIDC_ISSUER, dict(code=code, state=state))

        session_id = uuid4()
        username = aai_response["userinfo"]["name"]
        aai_id = aai_response["userinfo"]["sub"]

        session_data = SessionData(
            username=username,
            aai_state=state,
            aai_id=aai_id,
            session_uuid=str(uuid.uuid4()),
        )
```

michal-szostak (approved):

backend/app/routes/web/recommendation.py
Michal-Kolomanski: We should pass here the same AAI_UID which is associated with the user in the MP's DB. This ID is used to uniquely identify the user across all services. That being said, that has remained me that there is a bug causing duplication of users with the same AAI_UID. We should fix it in the new sprint.

Marcelinna (approved):

Issue #404, opened 2022-12-22
Search by long title problem


Marcelinna:
Service title:
SeaDataNet European Directory of the Initial OceanObserving Systems

Develop
![image](https://user-images.githubusercontent.com/59872096/209125676-fba253ea-38fe-4fe4-8e7e-e23d1eca27c1.png)

Production
![image](https://user-images.githubusercontent.com/59872096/209125647-c495d544-34ad-41af-9efa-4d9429af0f07.png)

tomaszandel:
In develop it looks fixed:
<img width="942" alt="Zrzut ekranu 2023-02-5 o 17 16 56" src="https://user-images.githubusercontent.com/11340871/216821235-f27ef51a-94d4-4125-a539-672c3511aea5.png">

Issue #397, opened 2022-12-16
Improve the search engine of the search service


agpul:
Input that we have: 
1. https://docs.google.com/spreadsheets/d/1LqWUNfL9dzjNd-34Kr29z9T1MS5icv0GX9p8wlWCjwA/edit#gid=0
2. Add the possibility of searching by providers/providers' acronym
3. Add the possibility of searching by two words together for eg. "EOSC Synergy" should find the first things related o both words

agpul:
Additional use cases:
https://docs.cyfronet.pl/display/FID/Search+Service+Workshop

tomaszandel:
please change providers field to Tokenized in schema and reload the data so we can search using Tokenizer @MIchalKolomanski

Issue #319, opened 2022-11-10
Multiple words search improvement


danielkryska:
**Possible cause**
SOLR queries have set up by default the wrong operator `OR` instead of `AND`.

**How to improve it?**
Add to each request `&q.op=AND`.

**Sources**
![image](https://user-images.githubusercontent.com/31220811/201025127-b265d77a-3980-4e45-8587-28262c96c84a.png)
https://stackoverflow.com/questions/24642900/solr-difficulty-getting-fuzzy-search-with-multiple-terms

![image](https://user-images.githubusercontent.com/31220811/201025182-fc9e2138-e3e5-40e1-bb43-8c033ff77236.png)
https://stackoverflow.com/questions/14226424/changing-the-default-operator-from-or-to-and-in-solr-magento-enterprise

Marcelinna:
duplicate https://github.com/cyfronet-fid/eosc-search-service/issues/404?

PR #435, opened 2023-02-06
feat: search improvements for #404 and #397 and #319

tomaszandel:
- search improvement
- tested on http://149.156.182.2:8983/solr and with NON facade recommender.

-goal: set KeywordTokenizer and add simple algo for KeywordTokenizer with ~2 maximum if more than 3 words with words size greater than 5. otherwise no fuzzy. case for ~1 also to be crafted. Important with KeywordTokenizer would be to add mm at 70-80%.

### DIFF
--- a/backend/app/routes/web/search.py
+++ b/backend/app/routes/web/search.py
@@ -3,7 +3,7 @@
 from json import JSONDecodeError
 
 from fastapi import APIRouter, Body, Depends, HTTPException, Query
-from httpx import AsyncClient
+from httpx import AsyncClient, TransportError
 
 from app.schemas.web import SearchRequest
 from app.solr.operations import search_dep
@@ -53,7 +53,7 @@ async def search_post(
                 cursor=cursor,
                 facets=request.facets,
             )
-        except Exception as e:
+        except TransportError as e:
             raise HTTPException(status_code=500, detail="Try again later") from e
     if response.is_error:
         try:
--- a/backend/app/routes/web/search.py
+++ b/backend/app/routes/web/search.py
@@ -18,7 +18,7 @@ router = APIRouter()
 async def search_post(
     collection: str = Query(..., description="Collection"),
     q: str = Query(..., description="Free-form query string"),
-    qf: list[str] = Query([], description="Query fields", example=["authors", "title"]),
+    qf: str = Query(..., description="Query fields"),
     fq: list[str] = Query(
         [],
         description="Filter query",
@@ -27,7 +27,7 @@ async def search_post(
     sort: list[str] = Query(
         [], description="Sort order", example=["description asc", "name desc"]
     ),
-    rows: int = Query(10, description="Row count", gte=0, le=100),
+    rows: int = Query(10, description="Row count", gte=3, le=100),
     cursor: str = Query("*", description="Cursor"),
     request: SearchRequest = Body(..., description="Request body"),
     search=Depends(search_dep),
@@ -53,7 +53,7 @@ async def search_post(
                 cursor=cursor,
                 facets=request.facets,
             )
-        except TransportError as e:
+        except Exception as e:
             raise HTTPException(status_code=500, detail="Try again later") from e
     if response.is_error:
         try:
--- a/backend/app/routes/web/search.py
+++ b/backend/app/routes/web/search.py
@@ -3,7 +3,7 @@
 from json import JSONDecodeError
 
 from fastapi import APIRouter, Body, Depends, HTTPException, Query
-from httpx import AsyncClient, TransportError
+from httpx import AsyncClient
 
 from app.schemas.web import SearchRequest
 from app.solr.operations import search_dep
--- a/transform/Pipfile
+++ b/transform/Pipfile
@@ -18,4 +18,3 @@ black = {extras = ["jupyter"], version = "*"}
 
 [requires]
 python_version = "3.10"
-python_full_version = "3.10.5"
--- a/transform/Pipfile.lock
+++ b/transform/Pipfile.lock
@@ -1,11 +1,10 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "b05676247bc1bee62f02110fb165a64718b360e4ad16a64c39a16bb1f9221d46"
+            "sha256": "7fa9536461fcb4880f7c56ec5fc6bb3dea620dd7cfda91fd123d8b1a4b79989a"
         },
         "pipfile-spec": 6,
         "requires": {
-            "python_full_version": "3.10.5",
             "python_version": "3.10"
         },
         "sources": [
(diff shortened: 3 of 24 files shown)


agpul:
Use case: Datasource: "MetalPDB: a database of metal-binding sites in 3D structures of biological macromolecules"
1. Type: "a database of metal-binding"
- metal-binding is changed to metalbinding and it is not found
- the word "a" causes a lot of problems because the search engine is looking for single "a"
image

agpul:
> Use case: Datasource: "MetalPDB: a database of metal-binding sites in 3D structures of biological macromolecules"
> 
> 1. Type: "a database of metal-binding"
> 
> * metal-binding is changed to metalbinding and it is not found
> * the word "a" causes a lot of problems because the search engine is looking for single "a"
>   image

maybe we should not take into consideration short words like "a", "of" etc... is it possible?

agpul:
Remove: 
image

agpul:
Just a question: 
image
What do you think, what should be more important: August x2 or "August 1964" - for me "August 1964":)

agpul:
> Just a question: image What do you think, what should be more important in this case: August x2 or "August 1964" ?- for me "August 1964":)

Or maybe we should add the support for ""? In case of such typing: 
image
we should have "August 1964" on the top.

eldenkat:
###

agpul:
> 

@eldenkat what is the problem here?

agpul:
I type: Sleeping site selection in two Asian viverrids: effects of predation risk, resource access and habitat characteristics"

why "habitat characteristics" is not highlighted in the title but is highlighted in the description?:)
image

agpul:
> > 
> 
> @eldenkat what is the problem here? Seems odd, no name, no title, three times the same position. What does it say to the user?

I think that it is not a scope of this issue because the search engine works properly - it is not the blame of the search engine it is the blame of OpenAire data, but @tomaszandel can confirm or reject:P

eldenkat:
Do author and title carry the same importance when searching by typed word? @agpul

eldenkat:
make the clear all button opaque?

eldenkat (changes requested):
Pagination does not work when results are listed on more than one page

eldenkat:
No highlights in trainings. Other categories shows higlights in the same search case.

eldenkat:
SS does not display a search result for the query from the first item in the list.

Marcelinna:
Why [Bouchaib Bahli] is higer than [Maksi Bouchaib]
image

Marcelinna:
I am not sure that autocomplete works fine
image

tomaszandel:
> Just a question: image What do you think, what should be more important in this case: August x2 or "August 1964" ?- for me "August 1964":)

Yes, this is why keywordTokenizer is probably better idea, it shoudl match whole sentences better,  but in some other cases it will lack search flexibility ;)

tomaszandel:
> > Use case: Datasource: "MetalPDB: a database of metal-binding sites in 3D structures of biological macromolecules"
> > 
> > 1. Type: "a database of metal-binding"
> > 
> > 
> > * metal-binding is changed to metalbinding and it is not found
> > * the word "a" causes a lot of problems because the search engine is looking for single "a"
> >   image
> 
> maybe we should not take into consideration short words like "a", "of" etc... is it possible?

Solr is returning that values properly, bug is on our side, we match the regex wrongly, thus it highlights every occurence, fix TODO

tomaszandel:
> > Just a question: image What do you think, what should be more important in this case: August x2 or "August 1964" ?- for me "August 1964":)
> 
> Or maybe we should add the support for ""? In case of such typing: image we should have "August 1964" on the top.

hmm in fact it is correct according to our setup, 2 matched names gets higher score, so its technically OK. TO fix it we should consider a separate call like you mention " " that looks for 100% whole sentence match. Or just field selection where search would look for

tomaszandel:
> > Use case: Datasource: "MetalPDB: a database of metal-binding sites in 3D structures of biological macromolecules"
> > 
> > 1. Type: "a database of metal-binding"
> > 
> > 
> > * metal-binding is changed to metalbinding and it is not found
> > * the word "a" causes a lot of problems because the search engine is looking for single "a"
> >   image
> 
> maybe we should not take into consideration short words like "a", "of" etc... is it possible?

- we can fix, we should make a list of separators we want to let in, when i remove - search works properly

tomaszandel:
> I type: Sleeping site selection in two Asian viverrids: effects of predation risk, resource access and habitat characteristics"
> 
> why "habitat characteristics" is not highlighted in the title but is highlighted in the description?:) image

This is interesting case, because solr indeed is not highlighting it in title, but its a bit more complicated how he chooses what and what not to highlight, we could fix it on our end as we are using regex match, however why solr is not highlighting it is a mystery :)

tomaszandel:
> Do author and title carry the same importance when searching by typed word? @agpul image image
> 
> which result should be higher? or should have the same weight

Yeah this is tricky, if want to locate by author names, author has to be boosted higher, otherwise simple names in many cases exists in titles, same with surnames. So its a tradeoff, ideally we should have a field selection in search that would fix everything

tomaszandel:
> No highlights in trainings. Other categories show higlights in the same search case.
> 
> 

yeah it seems its a case not handled on our end, trainings are not highlighed at all, its a valid bug, solr replies correctly here

tomaszandel:
> I am not sure that autocomplete works fine image

It is ok, remember that this is not really an autocomplete, because it searches title, descrition, authors, keywords and so on, if we want to have a real autocomplete for titles than we should search only in titles, or display 'autocomplete' tool differently, so we can see what is matching, here we show only titles, but matches could be in some other place, 

Zrzut ekranu 2023-02-7 o 23 14 29

Marcelinna:
I type "INDIGO Identity and Access Management (IAM)", when I am on page 1 
image
I type "INDIGO Identity and Access Management (IAM)", when I am on page 6
image

backend/app/routes/web/search.py
```diff
@@ -53,7 +53,7 @@ async def search_post(
                 cursor=cursor,
                 facets=request.facets,
             )
-        except TransportError as e:
+        except Exception as e:
```
Michal-Kolomanski, line 56: Too broad exception handling. If `TransportError` is not enough, let's add others

transform/Pipfile
```diff
@@ -18,4 +18,3 @@ black = {extras = ["jupyter"], version = "*"}
 
 [requires]
 python_version = "3.10"
-python_full_version = "3.10.5"
```
Michal-Kolomanski, line 21: Please revert those changes, they are not in the scope

transform/Pipfile.lock
```diff
@@ -1,11 +1,10 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "b05676247bc1bee62f02110fb165a64718b360e4ad16a64c39a16bb1f9221d46"
+            "sha256": "7fa9536461fcb4880f7c56ec5fc6bb3dea620dd7cfda91fd123d8b1a4b79989a"
```
Michal-Kolomanski, line 4: Please revert those changes, they are not in the scope

Michal-Kolomanski (approved):
LGTM

eldenkat:
Why a strict match doesn't go first?

eldenkat:
Highlights.
Entering "horizontal service", page 1-4 highlights word "service", page 6 - no higlights and than returning to the page one by one 5,4,3,2,1 - no highlights.

Marcelinna (approved):

eldenkat (approved):

PR #472, opened 2023-03-09
feat: [#450] sort by relevance

### LINKED ISSUE #450: Add "Sort by relevance" feature in SS
TBD

Michal-Kolomanski:
closes #450


michal-szostak (approved):
The code itself is ok. I'd move `routes/web/recommendation_utils` to something outside of the routes dir - maybe dependencies, or recommender. But it can stay as is for now.

Lack of tests is a bigger problem, but we can push this feature out and fix tests later. 

Lastly - final solution should be to have sort by relevance in the query endpoint, but as we discussed it can be a next step.

PR #473, opened 2023-03-09
461 add sort by functionality in the search service

### LINKED ISSUE #461: Add "sort by" functionality
According to the Figma design: https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/Marketplace-EOSC?node-id=609%3A7490&t=UOnUTAtnQd6kRuGM-0 add sort by functionality in the Search Service.

Acceptance criteria:
- [ ] As a User I can see products sorted by relevance  ("Relevancy")
- [ ] As a User I can see products sorted by "Date - Most recent" as a default
- [ ] As a User I can see products sorted by "Date - Least recent"
- [ ] As a User I can see products sorted by popularity - "Most popular"

Anok666:
Closes #461

### DIFF
--- a/ui/apps/ui/src/app/components/search-input/search-input.component.ts
+++ b/ui/apps/ui/src/app/components/search-input/search-input.component.ts
@@ -282,7 +282,7 @@ export class SearchInputComponent implements OnInit {
       queryParams: {
         q: sanitizeQuery(q) ?? '*',
       },
-      //queryParamsHandling: 'merge',
+      queryParamsHandling: 'merge',
     });
   }
 
--- a/ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
+++ b/ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
@@ -1,0 +1,103 @@
+import { Component, OnInit } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { Router } from '@angular/router';
+import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
+import { CustomRoute } from '@collections/services/custom-route.service';
+import { sanitizeQuery } from '@components/search-input/query.sanitizer';
+import { filter, switchMap } from 'rxjs';
+import {
+  DEFAULT_SORT,
+  isSortOption,
+  sortType,
+} from '@components/sort-by-functionality/sort-value.type';
+
+@UntilDestroy()
+@Component({
+  selector: 'ess-sort-by-functionality',
+  template: `<div class="col-sm-4 sort_container">
+    <label for="sorts" i18n>Sort By</label>
+    <select
+      [formControl]="selectedSortOptionControl"
+      id="sorts"
+      class="form-select"
+    >
+      <option value="dmr" i18n>Date - Most recent</option>
+      <option value="dlr" i18n>Date – Least recent</option>
+      <option value="mp" i18n>Most popular</option>
+      <option value="r" i18n>Relevancy</option>
+    </select>
+  </div>`,
+  styles: [
+    `
+      :host {
+        display: contents;
+      }
+
+      label {
+        width: 100px;
+        font-family: 'Inter';
+        font-style: normal;
+        font-weight: 400;
+        font-size: 14px;
+        line-height: 20px;
+        color: #4f4f4f;
+      }
+
+      select {
+        font-family: 'Inter';
+        font-style: normal;
+        font-weight: 500;
+        font-size: 14px;
+        line-height: 24px;
+        color: #101828;
+      }
+
+      select:focus {
+        -webkit-box-shadow: none;
+        box-shadow: none;
+        border: 1px solid #ced4da;
+      }
+
+      .sort_container {
+        width: 18rem;
+        display: flex;
+        align-items: center;
+      }
+    `,
+  ],
+})
+export class SortByFunctionalityComponent implements OnInit {
+  public selectedSortOptionControl: FormControl<sortType> =
+    new FormControl<sortType>(DEFAULT_SORT, { nonNullable: true });
+
+  constructor(private _customRoute: CustomRoute, private _router: Router) {}
+
+  ngOnInit() {
+    this._customRoute.sortUI$
+      .pipe(
+        untilDestroyed(this), // TODO: event triggered on clear or change search criteria
+        filter((sortUI) => sortUI !== this.selectedSortOptionControl.value)
+      )
+      .subscribe((sortUI) => {
+        this.selectedSortOptionControl.setValue(
+          isSortOption(sortUI) ? sortUI : DEFAULT_SORT,
+          { emitEvent: false }
+        );
+      });
+    this.selectedSortOptionControl.valueChanges
+      .pipe(
+        untilDestroyed(this),
+        switchMap((value) => this.updateQueryParams(value))
+      )
+      .subscribe();
+  }
+
+  async updateQueryParams(sortOption: sortType) {
+    await this._router.navigate([], {
+      queryParams: {
+        sortUI: sanitizeQuery(sortOption) ?? DEFAULT_SORT,
+      },
+      queryParamsHandling: 'merge',
+    });
+  }
+}
--- a/ui/apps/ui/src/app/collections/services/custom-route.service.ts
+++ b/ui/apps/ui/src/app/collections/services/custom-route.service.ts
@@ -7,10 +7,12 @@ import { queryParamsMapFrom } from '@collections/services/custom-route.utils';
 import { toArray } from '@collections/filters-serializers/utils';
 import { serializeAll } from '@collections/filters-serializers/filters-serializers.utils';
 import { FiltersConfigsRepository } from '@collections/repositories/filters-configs.repository';
+import { DEFAULT_SORT } from '@components/sort-by-functionality/sort-value.type';
 
 const DEFAULT_PARAMS = {
   collection: null,
   q: '*',
+  sortUI: DEFAULT_SORT,
   fq: [],
   cursor: '*',
   sort: [],
@@ -35,6 +37,7 @@ export class CustomRoute {
     map((collection) => collection as string)
   );
   readonly q$ = this._store$.pipe(select(({ q }) => q));
+  readonly sortUI$ = this._store$.pipe(select(({ sortUI }) => sortUI));
   readonly fqMap$ = this._store$.pipe(
     select(({ fq }) =>
       serializeAll(
--- a/ui/apps/ui/src/app/collections/services/custom-route.type.ts
+++ b/ui/apps/ui/src/app/collections/services/custom-route.type.ts
@@ -1,6 +1,9 @@
+import { sortType } from '@components/sort-by-functionality/sort-value.type';
+
 export type paramType =
   | string
   | string[]
+  | sortType
   | undefined
   | null
   | number
@@ -8,6 +11,7 @@ export type paramType =
 export interface ICustomRouteProps {
   collection: string | null;
   q: string;
+  sortUI: sortType;
   fq: string[];
   cursor: string;
   sort: string[];
--- a/ui/apps/ui/src/app/components/page-header/page-header.component.ts
+++ b/ui/apps/ui/src/app/components/page-header/page-header.component.ts
@@ -8,8 +8,10 @@ import { Router } from '@angular/router';
   selector: 'ess-page-header',
   template: `
     <div id="container" class="page-heading">
-      <span id="results-count" i18n>{{ resultsCount }} search results</span> in
-      {{ (activeNavConfig$ | async)?.title }}
+      <div>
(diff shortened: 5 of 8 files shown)


ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,146 @@
...
+        display: flex;
+        align-items: center;
+      }
+    `,
+  ],
+})
+export class SortByFunctionalityComponent implements OnInit {
+  public selectedSortOption: string = DefaultSortValueEnum.DEFAULT;
+
+  public readonly sortOptions: Array<ISortOption> = [
```
michal-szostak, line 77: Instead of having this array in the component I would have sort option interface which would couple caption and value. Here you have separate enums `SortOptionValuesEnum` and `DefaultSortValueEnum`. I think that having those values specified in the template itself (`<option></...`) would be simpler. This is not something that is fetched from server or calculated. So having it in template would be simpler.

ui/apps/ui/src/app/components/sort-by-functionality/sort-option-values.enum.ts
```diff
@@ -0,0 +1,6 @@
+export enum SortOptionValuesEnum {
+  RELEVANCY = 'Relevancy',
```
michal-szostak, line 2: If your defining user readable text you probably should use `$localize`. But I do not think that storing captions for end user in the enum is the best idea.

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,146 @@
...
+    private _customRoute: CustomRoute,
+    private _navConfigsRepository: NavConfigsRepository,
+    private _router: Router
+  ) {}
+
+  ngOnInit() {
+    this._customRoute.s$
+      .pipe(
+        untilDestroyed(this),
+        filter((s: string) => s !== DefaultSortValueEnum.DEFAULT)
```
michal-szostak, line 112: I don't think that it would work as expected. If user changes select to other value and then tries to go back to `DEFAULT` the option would not register.

Additionally this is something that is duplicated mostly by `updateQueryParams`.

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,146 @@
...
+    let searchQueryParamsQ;
+
+    this._customRoute.q$
+      .pipe(
+        untilDestroyed(this),
+        filter((q: string) => q !== '*')
+      )
+      .subscribe((q) => {
+        searchQueryParamsQ = q;
+      });
```
michal-szostak, line 129: This subscription will cause memory leak and will be spawned for every call of the updateQueryParams, which is probably not something we want.
Also spawning two async functions (first this and then navigate) is probably not the best idea. I would either chain it with `.navigate` or wrap it in await.

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,146 @@
...
+@UntilDestroy()
+@Component({
+  selector: 'ess-sort-by-functionality',
+  template: `<div class="col-sm-4 sort_container">
+    <label for="sorts">Sort By</label>
+    <select
+      [(ngModel)]="selectedSortOption"
+      #sorts
+      id="sorts"
+      label="Sort by"
```
michal-szostak, line 26: `i18n-label`

.gitignore
```diff
@@ -73,6 +73,7 @@ typings/
 .env
 .env.test
 .env*.local
+.env.template
```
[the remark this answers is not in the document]
Anok666: entry deleted

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,146 @@
...
+      class="form-select"
+      (change)="this.onSelected(sorts.value)"
+    >
+      <option *ngFor="let sortBy of sortOptions" value="{{ sortBy.value }}">
```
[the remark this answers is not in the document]
Anok666: fixed

ui/apps/ui/src/app/components/sort-by-functionality/sort-option-values.enum.ts
Anok666: i18n labels added

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
Anok666: I solved differently

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,105 @@
...
+      .sort_container {
+        width: 18rem;
+        display: flex;
+        align-items: center;
+      }
+    `,
+  ],
+})
+export class SortByFunctionalityComponent implements OnInit {
+  public selectedSortOptionControl: FormControl = new FormControl(DEFAULT_SORT);
```
michal-szostak, line 66: FormControl's in Angular 14+ are typed. It would be good to define types elsewhere (for example in `sort-value-default.ts` e.g.:

```typescript
// sort-value-default.ts
export type sortType = 'r' | 'dmr' | 'dlr' | 'mp';
```

```typescript
// sort-by-functionality.component.ts
public selectedSortOptionControl: FormControl<sortType> =
    new FormControl<sortType>(DEFAULT_SORT, { nonNullable: true });
```

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        this.selectedSortOptionControl.setValue(s, { emitEvent: false });
+      });
+    this.selectedSortOptionControl.valueChanges
+      .pipe(untilDestroyed(this))
+      .subscribe((value) => {
+        this.updateSortOption(value);
+      });
+  }
+
+  async updateSortOption(sortOption: string) {
```
michal-szostak, line 86: And `sortOption` could be typed to be `(sortOption: sortType)`

ui/apps/ui/src/app/components/sort-by-functionality/sort-value-default.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_SORT = 'dlr';
```
michal-szostak, line 1: this const could also be typed to defined `sortType`:

```typescript
// sort-value-default.ts
export type sortType = 'r' | 'dmr' | 'dlr' | 'mp';
export const DEFAULT_SORT: sortType = 'dlr';
```

Also - I'd consider changing name of the file to something along the lines of `sort-value.types.ts`

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,105 @@
...
+        untilDestroyed(this), // TODO: event triggered on clear or change search criteria
+        filter((s) => s !== this.selectedSortOptionControl.value)
+      )
+      .subscribe((s) => {
+        this.selectedSortOptionControl.setValue(s, { emitEvent: false });
+      });
+    this.selectedSortOptionControl.valueChanges
+      .pipe(untilDestroyed(this))
+      .subscribe((value) => {
+        this.updateSortOption(value);
```
michal-szostak, line 82: Here you call async function from within subscribe - it is generally not the best practice.
I'd suggest following way of handling it:

```typescript
    this.selectedSortOptionControl.valueChanges
      .pipe(
        untilDestroyed(this),
        switchMap((value) => this.updateSortOption(value))
      )
      .subscribe();
```

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  ngOnInit() {
+    this._customRoute.s$
+      .pipe(
+        untilDestroyed(this), // TODO: event triggered on clear or change search criteria
+        filter((s) => s !== this.selectedSortOptionControl.value)
+      )
+      .subscribe((s) => {
+        this.selectedSortOptionControl.setValue(s, { emitEvent: false });
+      });
```
michal-szostak, line 78: Here some typing could be used if we have form control which is strongly typed:

```typescript
    this._customRoute.s$
      .pipe(
        untilDestroyed(this), // TODO: event triggered on clear or change search criteria
        filter((s) => s !== this.selectedSortOptionControl.value)
      )
      .subscribe((s) => {
        this.selectedSortOptionControl.setValue(
          isSortOption(s) ? s : DEFAULT_SORT,
          { emitEvent: false }
        );
      });

```

The `isSortOption` is guard option which could be used to make sure that `s` is a given subtype. (It could also be placed in `sort-value-default.ts`)

```typescript
export const DEFAULT_SORT = 'dlr';
export type sortType = 'r' | 'dmr' | 'dlr' | 'mp';


export function isSortOption(obj: string): obj is sortType {
  return ['dlr', 'dmr', 'mp', 'r'].includes(obj);
}

```

michal-szostak (changes requested):

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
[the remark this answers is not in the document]
Anok666: I had a problem with calling async await from subscribe / ngOnInit. fixed.

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  async updateQueryParams(sortOption: string) {
+    this._customRoute.params$
+      .pipe(
+        take(1),
+        filter((paramsS) => paramsS.s !== sortOption)
+      )
+      .subscribe((params) => {
+        this._router.navigate([], {
```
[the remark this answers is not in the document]
Anok666: I thought that there must be a better option and a method to do it. fixed.

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
Anok666: Nice. fixed.

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
Anok666: I like this approach. done.

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
Anok666: done.

ui/apps/ui/src/app/components/sort-by-functionality/sort-by-functionality.component.ts
Anok666: Nice. done.

michal-szostak (approved):

Marcelinna:
Can you display the value in the same order as in Figma ?
FIGMA
image
```

Date - Most recent
Date - Last recent
Most popular
Relevance

```

PR
image

Marcelinna:
- [ ] QA
- [ ] DEV

image

Most recent should be as default

Marcelinna:
- [ ] QA
- [ ] DEV
1. Go to to SS
2. Change Sort by to "Most popular"
3. Type data to search bar 
4. Sort by is changed to "Data - Least recent"

1. Go to to SS
2. Type data to search bar - "data"
3. Change Sort by to "Most popular" 
image
5. Click "Clear" in serach bar 
image
7. Sort by is changed to "Data - Least recent"

ui/apps/ui/src/app/components/search-input/search-input.component.ts
```diff
@@ -281,7 +281,7 @@ export class SearchInputComponent implements OnInit {
       queryParams: {
         q: sanitizeQuery(q) ?? '*',
       },
-      //queryParamsHandling: 'merge',
+      queryParamsHandling: 'merge',
```
michal-szostak, line 284: I'm not 100% if this is behavior we desire. This might break some things.

Anok666: I consulted and this change was tested.

Issue #480, opened 2023-03-07
Add Interoperability Guidliness inthe user interface in SS


agpul:
We need to add the Interoperability Guidliness in the User Interface of the Search Service
Acceptance criteria:
- [ ] As a User I can see the dedicated space for Interoperability Guidelines on the main page of the search service [1]
- [ ] As a User I can read the dedicated text about the Interoperability Guidelines on the main page of the SS according to [1]
- [ ] As a User I can click on the link under 
![image](https://user-images.githubusercontent.com/43000481/224690980-9791d918-d5a6-4930-a879-4d7b3868e607.png)
- [ ] As a User I can go to the dedicated page for the INteroperability guidelines in the search service for eg. /interoperability-guidelines
- [ ] As a User I can see a page with IG like [2]
- [ ] As a User I can come back to the search  using the "Go to search" button
- [ ] As a User I can see the information about each Guideline as follow:
![image](https://user-images.githubusercontent.com/43000481/224691848-fe476e16-816a-4f50-a681-615e53a983bd.png)
where:
- EOSC Guidline Type
- "EOSC Resource Catalog: Architecture..." is the title
- "Eosc Research..." is the description
- logo -> not now, we don't have such data
- "EOSC Future" is the organization's abbreviation
- "European Open Science Cloud Future" is the full name of the organization
- [ ]  As a User I can see the "Learn more...." section according to [2] where links lead to from top to bottom:
- https://eosc-portal.eu/eosc-interoperability-framework/about-eosc-interoperability-framework-governance-eosc-if
- https://eosc-portal.eu/eosc-interoperability-framework/eiab-and-eiac-charter
- https://eosc-portal.eu/eosc-interoperability-framework/eosc-if-templates-and-information
Figma:
[1] https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/Marketplace-EOSC?node-id=440-3486&t=sPFSFvby7JUO9rZ6-0
[2] : https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/Marketplace-EOSC?node-id=1618%3A14390&t=Ceae7xz8A8J1nGcI-0#386402438

Anok666:
The functionality will be implemented elsewhere. Change of logic after stakeholders' review.

PR #498, opened 2023-03-22
feat: #480 interoperability guidelines added to user interface of SS

Anok666:
The Interoperability Guidelines Component was added to user interface of Search Service.
Simple text with background color and link in headline to direct to Interoperability Guidelines Page.


Anok666:
The functionality will be implemented elsewhere. Change of logic after stakeholders' review.

Issue #499, opened 2023-03-23
Add Interoperability Guidelines Page to SS


Anok666:
Interoperability Guidelines Page:

A view that shows the whole collection of guidelines from json mock.

Anok666:
The functionality will be implemented elsewhere. Change of logic after stakeholders' review.

PR #500, opened 2023-03-24
feat: #499 interoperability guidelines page added. work in progress

Anok666:
interoperability guidelines page added. work in progress


Anok666:
The functionality will be implemented elsewhere. Change of logic after stakeholders' review.

PR #533, opened 2023-04-12
Back to search redirects from search

tomaszandel:
- alow users in other services to have a custom go back to search button


Marcelinna (approved):

version.txt
```diff
@@ -1 +1 @@
-2.1.0
+2.1.1
```
Michal-Kolomanski, line 1: Not sure if that also should be provided before a release itself

PR #547, opened 2023-04-19
fix: [#428] Recommendation panel user action root content fixed

### LINKED ISSUE #428: After clicking on the recommendation panel, type is other instead of recommendation_panel
When a user clicks on the recommendation panel type should be recommendation_panel

>  "root": {
>       "type": "recommendation_panel",
>       "panel_id": "v1",
>       "resource_id": "123",
>       "resource_type": "service", 
>     }

wujuu:
The issue was that in the recommendation panel, there was no flag specifying that the user clicked the actual recommendation panel.

Closes #428


Marcelinna (approved):

Michal-Kolomanski (approved):
Tested :heavy_check_mark:

Issue #541, opened 2023-04-13
Design the mechanism of merging user_actions from SS and MP into a single user session


Michal-Kolomanski:
- [ ] Design the mechanism for merging user actions from Search Service, and Marketplace (later on User Dashboard) into a single user journey. Please document the design somewhere and share it. Consult/work with that design with @michal-szostak, @Michal-Kolomanski

Key aims:
- we need to be able to construct a full user journey across SS, MP, (UD later)
- recommendations should be linkable to user actions
- source/target should be consistent across user action events, no matter whether a user got redirected to another service. Watch out for [4. Deprecation of marketplace service list and Inconsistencies in new user action path patterns introduced](https://wiki.eoscfuture.eu/display/EOSCF/Issues+and+inconsistencies+in+incoming+data). This inconsistency is about: SS tends to put a whole resource URL link into source/target while in MP there is a path. That should be fixed in a separate ticket. 

Keynotes:
- currently, there are no user actions in UD so we can simplify the problem by excluding UD from the equation, but it will be great to be able to add user cations easily in UD later on - so please keep UD in mind.
- domain: [eosc-portal.eu](http://eosc-portal.eu/)
- subdomain: [marketplace.eosc-portal.eu](http://marketplace.eosc-portal.eu/) and [my.eosc-portal.eu](http://my.eosc-portal.eu/)
- subsubdomain: [search.marketplace.eosc-portal.eu](http://search.marketplace.eosc-portal.eu/)
- subdomains should be able to get/pass some information about each other right? That should help
- user can start a journey in SS, MP (only direct link to the resource, there is no service catalog in MP anymore), UD
- recommendations are shown in SS and UD

PR #613, opened 2023-05-21
feat: [#542] Forward user action details to other services

### LINKED ISSUE #542: Develop the mechanism of merging user_actions from various services
Development of a design prepared in: https://github.com/cyfronet-fid/eosc-search-service/issues/541

wujuu:
Closes #542 and #541

### DIFF
--- a/backend/app/routes/web/user_actions.py
+++ b/backend/app/routes/web/user_actions.py
@@ -57,28 +57,27 @@ async def register_navigation_user_action(
     if url.startswith("/"):
         url = UI_BASE_URL + url
 
-    url_params = (
-        url
-        + ("&" if "?id" in url else "?")
-        + "return_path="
-        + return_path
-        + "&search_params="
-        + search_params
-    )
-
-    response = RedirectResponse(status_code=303, url=url_params)
+    target_id = uuid.uuid4()
 
     try:
         cookie(request)
         session = await verifier(request)
-    except HTTPException:
-        session_id = uuid.uuid4()
 
-        session = SessionData(
-            username=None, aai_state=None, aai_id="", session_uuid=str(uuid.uuid4())
+        response = await _create_redirect_response(
+            return_path, search_params, session.session_uuid, str(target_id), url
         )
-        await backend.create(session_id, session)
-        cookie.attach_to_response(response, session_id)
+    except HTTPException:
+        cookie_session_id = uuid.uuid4()
+        new_session_uuid = uuid.uuid4()
+        session = SessionData(
+            username=None, aai_state=None, aai_id="", session_uuid=str(new_session_uuid)
+        )
+        await backend.create(cookie_session_id, session)
+
+        response = await _create_redirect_response(
+            return_path, search_params, str(new_session_uuid), str(target_id), url
+        )
+        cookie.attach_to_response(response, cookie_session_id)
 
     if not client:
         logger.debug("No mqtt client, user action not sent")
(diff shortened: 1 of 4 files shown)


Michal-Kolomanski:
Please fix the tests. They seemed to be related to UA

Marcelinna:
@wujuu write how to test.

wujuu:
@Marcelinna we want to connect to the databus and then check if user actions originating from SS are being linked properly to user actions originating from MP.

1. Visit SS
2. Go to 'services'
3. Click a service (note the user action, target.visit_id)
4. [Be redirected to MP]
5. Click something on the service page in MP (note the user action. source.visit_id)

Compare the 2 visit_ids. They should be the same.

Notes:
1. When you click "back" in the browser or "Go to search" button the user action chain should be broken. This is expected.
2. User actions on the SS itself do not chain - this is for the future, but we only redirect the user for now so it should be fine.


About the testing environment - it will require local setup (cannot test on staging instances), becase it requires the updates from  https://github.com/cyfronet-fid/marketplace/pull/2990. This is how I did it:
1. Setup two local instances for SS and MP locally using `542-add-user-actions-forward` branch in SS and `2989-allow-ua-to-be-sent-from-others` in MP.
2. Connect to local STOMP server - where the user actions from local SS and MP instances are pushing user actions.
3. Subscribe to `/topic/user_actions` and print and incoming user actions.

ui/apps/ui/src/app/collections/services/redirect.service.ts
```diff
@@ -20,7 +20,7 @@ export class RedirectService {
     }
 
     const encodedPv = encodeURIComponent(
-      'search/' + this._customRoute.collection()
+      '/search/' + this._customRoute.collection()
```
michal-szostak, line 23: Forward slash here should not be added. We already informed our partners that this interface would not change (also it would require changes in MP)

michal-szostak (changes requested):

ui/apps/ui/src/app/collections/services/redirect.service.ts
wujuu: done

backend/app/routes/web/user_actions.py
```diff
@@ -56,13 +56,20 @@ async def register_navigation_user_action(
...
         url
         + ("&" if "?id" in url else "?")
         + "return_path="
         + return_path
         + "&search_params="
         + search_params
+        + "&source_id="
+        + target_id
+        + "&client_uid="
+        + session_uuid
```
tomaszandel, line 72: Why we are sending session uuid in url ?

michal-szostak (approved):

Michal-Kolomanski (approved):
@tomaszandel Here we also have to extract proper aai_uid (not create our own). How it should be done after your changes?

Michal-Kolomanski (commented):
I tried to test this locally but the env is hard to set up. Local SS redirects links to the production instance of MP. Furthermore, the actual aai_uid of a user has to be passed (otherwise it is not possible to identify a user). Sending aai_uid via params in a request is acceptable @michal-szostak, @tomaszandel?

wujuu:
Ok, just to be clear - *right now* in Marketplace "client_uid" is not tied to aai_id of the user in any way. It's just a unique identificator of a user browser session, that can later be used to chain user action together.

In Search Service the "client_uid" field in the user action payload is taken from the browser session (session_uuid) that can but doesn't have to tied to the actual aai_id. To implement the cross-site user action continuity I'm passing the mentioned session_uuid in the query params. This is *not* sensitive data. 

Some unresolved isses lie in the previous flow of the session creation which I have adjusted slightly but it still needs work like @tomaszandel mentioned.

PR #623, opened 2023-05-25
Mobile view fix

jarekzet:
Added mods for mobile version of search results

#622

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "eosc-search-service",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {}
+}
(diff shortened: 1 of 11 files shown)


agpul:
image
Add "Hide filters" button (the second one) under all filters.

agpul:
I can't see "Login" possibility

Marcelinna:
@aleksandrahojszyk @agpul it is ok, or we should make mobile view at higher resolution ?
image

Marcelinna:
Guideines view 
image

aleksandrahojszyk:
> @aleksandrahojszyk @agpul it is ok, or we should make mobile view at higher resolution ? image

I do agree to make mobile view starting from higher resolution

Marcelinna (approved):

Marcelinna (approved):

PR #637, opened 2023-06-01
feat: [#636] Enable sort by date/popularity/relevance

### LINKED ISSUE #636: Sort by date, pupularity

Michal-Kolomanski:
closes #636

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,8 +125,6 @@ See docker-compose.yml for components.
   > Client id used to identify databus (jms) client
 - `RECOMMENDER_ENDPOINT`
   > Recommender endpoint (default http://localhost:8081/recommendations)
-- `MAX_ITEMS_SORT_BY_RELEVANCE`
-  > The maximum number of resources for sort by relevance (default 50)
 - `MARKETPLACE_BASE_URL`
   > marketplace base url (used to generate links back to MP) (default https://marketplace.eosc-portal.eu)
 - `LOG_LEVEL`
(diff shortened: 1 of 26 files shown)


Marcelinna:
why do we have relevancy for services but for data sources no
image
image

Marcelinna:
is there any difference between most popular and relevance, it returns the same ?

Marcelinna:
Step to reproduce
1. go to https://search-3.docker-fid.grid.cyf-kr.edu.pl/search/all?q=*
2. Go to Publications
3. Set Sort by Relavance
image
4. Go to Data
5. See problem - Publications are displayed
image
6. Go to Services
7. See problem - Publications are displayed
image

Marcelinna:
When I refresh site I often see this 
image

Michal-Kolomanski:
> why do we have relevancy for services but for data sources no? becasue od UA?
> 
> image image

There is no sort by relevance for data sources / bundles / guidelines - the recommender is not handling that. It was also decided to do not to enable this in all catalogs

Michal-Kolomanski:
@agpul @Marcelinna For me, chosen filtering should stay the same as it is set during a collection switching. Currently, when you switch collections, filtering is set on default sorting (which is popularity)

Michal-Kolomanski:
@michal-szostak we are missing here loading logo. Could you describe how its loaded? Any ideas why it is the case?

Michal-Kolomanski:
aai_uid needs to be also added. @tomaszandel consult

Marcelinna (approved):

Issue #515, opened 2023-03-31
Redesign "Providers" in SS 🔥


agpul:
Like IG

https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/Marketplace-EOSC?type=design&node-id=3049%3A17709&mode=design&t=CAe5MLx9x0wHO8VV-1

aleksandrahojszyk:
https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/Marketplace-EOSC?type=design&node-id=3049%3A17709&mode=design&t=CAe5MLx9x0wHO8VV-1

PR #673, opened 2023-06-16
feat: #515 call provider

Anok666:
https://github.com/cyfronet-fid/eosc-search-service/issues/515

Acceptance criteria:

 As a User I want to see the link to the Providers' tab on the main page of the Search Service (right box) [1]
 As a User I want to see the site with the list of Providers with applied filters as [1] where each box has information as in design [1]
 As a user I want to search through providers using search in the tab (no providers in all catalog)
 As a User after clicking on the Provider name that redirects me to Providers in MP
Figma;
[1] https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/Marketplace-EOSC?type=design&node-id=3049%3A17709&t=Mg6BhaTTD7enweyA-1


Marcelinna:
When I click on the provider in autocomplete list
image
image

Marcelinna:
image

Marcelinna:
When I click show more I don't see all description 
image
image

Marcelinna:
@Anok666 
image

eldenkat (approved):

ui/apps/ui/src/app/components/collections-navigation/collections-navigation.component.ts
```diff
@@ -9,19 +9,21 @@ import { toNavigationLink } from './utils';
...
-            paths: 'exact',
-            matrixParams: 'ignored',
-            fragment: 'ignored'
-          }"
-          routerLinkActive="active"
-          [queryParams]="{ q: (q$ | async) }"
-          >{{ link.label }}</a
+        <ng-container *ngFor="let link of navigationLinks">
+          <a
+            *ngIf="link.label.toLowerCase() !== 'providers'"
```
michal-szostak, line 14: I don't like conditioning generic component based on the link's label name.

ui/apps/ui/src/app/components/recommendations/recommendations.component.ts
```diff
@@ -78,4 +83,8 @@ export class RecommendationsComponent implements OnInit {
       filtersConfigs
     );
   }
+
+  _joinDescription(desc: string | string[]): string {
```
michal-szostak, line 87: This function is public - it should not have `_` prefix. Additionally, it would be better to also create pipe so that template does not call it on every redraw cycle.

ui/apps/ui/src/app/components/results-with-pagination/result.component.ts
```diff
@@ -31,25 +31,43 @@ import { IOffer } from '@collections/data/bundles/bundle.model';
...
+          <ess-url-title
+            *ngIf="type.value === 'provider'"
+            [ngClass]="{ sub_title: type.value === 'provider' }"
+            [title]="abbreviation"
+            [highlight]="highlightsreal['abbreviation'] ?? []"
+            [url]="redirectService.internalUrl(validUrl, id, type.value, '')"
+            style="margin-bottom: -10px;"
+          >
+          </ess-url-title>
+        </div>
```
michal-szostak, line 69: Again - I'm not a fan of switch style templates. But this template already had bundles as a special condition, so it's passable. But will need refactoring in the future

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
```diff
@@ -37,29 +37,44 @@ import { Router } from '@angular/router';
...
       "
-      ><ess-pagination
-        [paginationData]="$any(paginationData$ | async)"
-        [loading]="(isLoading$ | async) ?? false"
-        (activePageChange)="pageNr$.next($event)"
-      ></ess-pagination>
-      <ess-result
-        class="results"
+    >
+      <ng-container
```
michal-szostak, line 41: You accidentally cut top pagination, which was introduced earlier. Please put it back.

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
```diff
@@ -37,29 +37,44 @@ import { Router } from '@angular/router';
...
-        [downloads]="result.downloads"
-        [accessRight]="result.accessRight"
-        [date]="result.date"
-        [highlights]="highlights[result.id] ?? {}"
-        [offers]="result.offers ?? []"
-      ></ess-result>
+      >
+        <ess-result
+          class="results"
+          *ngIf="!result.type.value.toString().includes('provider')"
```
michal-szostak, line 46: This is probably more complicated than it needs to be.

Also in this case either use `if; else` or `switch`. Creating opposite ifs can lead to bugs where both (or none) cases are displayed.

ui/apps/ui/src/app/components/right-menu/right-menu.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'ess-right-menu',
+  template: `<div>
+    <a [queryParams]="{ q: '*' }" [routerLink]="'/search/provider'"
+      >Providers -></a
+    >
+  </div>`,
+  styles: [``],
```
michal-szostak, line 10: If component does not have styles this part can be skipped

ui/apps/ui/src/app/components/search-input/search-input.component.ts
```diff
@@ -271,9 +271,10 @@ export class SearchInputComponent implements OnInit {
             .pipe(untilDestroyed(this))
         )
       )
-      .subscribe(
-        (suggestedResults) => (this.suggestedResults = suggestedResults)
-      );
+      .subscribe((suggestedResults) => {
+        this.suggestedResults = suggestedResults;
+        console.log(this.suggestedResults);
```
michal-szostak, line 276: Is this console log required? Or an artifact from debugging?

ui/apps/ui/src/app/dictionary/translateDictionaryValue.ts
```diff
@@ -74,7 +74,7 @@ export function translateDictionaryValue(
       return trainingDomainDictionary[valueType] || value;
       break;
     case DICTIONARY_TYPE_FOR_PIPE.GUIDELINE_PROVIDER:
-      return cleanGuidelineProvider(valueType);
+      return trainingDomainDictionary[valueType] || value; // cleanGuidelineProvider(valueType);
```
michal-szostak, line 77: If `cleanGuidelineProvider` is no longer required maybe the function itself should be removed?

michal-szostak (changes requested):
I have some general issues (like adding `rightMenu` flag to all collections or extensive use of `if` in templates. But given time constraints and the fact that similar code has already been present in the codebase I'm giving an approve. If possible some minor issues might be corrected, but I don't see anything that would prevent merging. 

The only thing that must be fixed is cutting top pagination.

ui/apps/ui/src/app/components/collections-navigation/collections-navigation.component.ts
Anok666: done. I moved the exclusion of providers from the top menu to the nav config registry and applied the filter on the result

ui/apps/ui/src/app/components/recommendations/recommendations.component.ts
Anok666: I changed to pipe

ui/apps/ui/src/app/components/results-with-pagination/result.component.ts
Anok666: I used the else if template

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
Anok666: yes, im sorry :). i restored

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
Anok666: I used the else if template

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
```diff
@@ -177,4 +192,9 @@ export class ResultsWithPaginationComponent implements OnInit {
     this._prevParamsWithoutCursor = paramsWithoutCursor;
     return shouldInitPagination;
   }
+
+  _joinDescription(desc: string | string[]): string {
```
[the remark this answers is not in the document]
Anok666: I changed to pipe

ui/apps/ui/src/app/components/right-menu/right-menu.component.ts
Anok666: styles added

ui/apps/ui/src/app/components/search-input/search-input.component.ts
Anok666: I was looking for something for Maria and forgot about it

ui/apps/ui/src/app/dictionary/translateDictionaryValue.ts
Anok666: done

michal-szostak (approved):

Marcelinna (approved):

Issue #395, opened 2022-12-15
Suggestions / Autocomplete endpoint


michal-szostak:
Currently front end stitches main search bar autocomplete from many requests. This should be hidden behind single API endpoint.

The scope of this issue:

* create new endpoint `/api/web/search-suggestions`. The endpoint should accept the same parameters as `search-results` with exception of: (will be filled). Additionally `results_per_collection=int` should be parsed to return given number of results per collection. (validate maximum - less than 10)
* The response should have following format
```
{
"<collection_name>": [
{
// collection entity as returned from solr
}, ...
]
}
```

PR #713, opened 2023-07-06
refactor: [#395] split suggestions on the backend side

maria-j-k:


maria-j-k:
Acceptance criteria:
example request: 
`http://127.0.0.1:8000/api/web/search-suggestions?q=house&qf=title%5E100%20author_names_tg%5E120%20description%5E10%20keywords_tg%5E10%20tag_list_tg%5E10&collection=all_collection&results_per_collection=3`
The endpoint should accept specific collection or "all_collection".
The endpoint should return the json with collection names as keys and results if found. For the collection with no result, the value should be an empty list.

Marcelinna (approved):

backend/tests/app/routes/test_suggestions.py
```diff
@@ -0,0 +1,101 @@
+# pylint: disable=missing-module-docstring,missing-function-docstring
+from unittest.mock import ANY, AsyncMock
+
+import pytest
+from fastapi import FastAPI
+from httpx import AsyncClient
+from starlette import status
+
+URL = "web:post-search-suggestions"
```
michal-szostak, line 9: First, this is not an `URL` - calling `app.url_path_for(URL)` is a little bit confusing.
Secondly it is a little bit contentious topic, but I believe, that it's better to specify URLs for tests without using resolver (`app.url_path_for`) In the end URL is part of the contract, so it should be defined and unchanging - the same as result and payload. Doing it this way will de facto exclude it from tests.

This is the standard of "do as I say, not as I do" as I understand that all of our tests do this. But it's more a convention that should change.

We can decide to retain `app.url_path_for` for now, but at least changing the name of the variable from `URL` to `SEARCH_SUGGESTION_PATH` would be advisable.

backend/tests/app/routes/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+    mock_json = json.loads(read_file_contents(file))
+    mock_return = Mock()
+    mock_return.is_error = False
+    mock_return.json = Mock(return_value=mock_json)
+    return create_autospec(search, return_value=mock_return)
+
+
+def read_file_contents(file: str):
+    with open(f"{os.path.dirname(__file__)}/{file}", "r", encoding="utf-8") as f:
+        return f.read()
```
michal-szostak, line 31: I would suggest moving `json.loads` inside of this function and calling it `read_file_as_json`.
Additionally I strongly advice using `pathlib` module instead of `os.path` module. `Path` is a much more Object Oriented way of doing things with file paths: (In this short case the code written would totally break on windows (which in our case is probably acceptable, but it is still something good to avoid as a basic rule - rember - not all computer systems use `/` as a separator in file systems)

```
from pathlib import Path
with open(Path(__file__).parent / file, "r", encoding="utf-8") as f:
        return f.read()

```

Also  - encoding is usually understood and default is `"utf-8"` - but being verbose is not wrong.

michal-szostak (changes requested):

backend/app/routes/web/search_suggestions.py
```diff
@@ -0,0 +1,112 @@
...
+    fq: list[str] = Query(
+        [],
+        description="Filter query",
+        example=["journal:Geonomos", 'journal:"Solar Energy"'],
+    ),
+    results_per_collection: int = Query(
+        3, description="Row count per collection", gte=3, lt=10
+    ),
+    search=Depends(search_dep),
+):
```
michal-szostak, line 80: It would be good to type return type, especially that this function returns tuple, which can be pretty tricky sometimes.

backend/tests/app/routes/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+@pytest.fixture
+def mock_post_search(app: FastAPI) -> AsyncMock:
+    mock_search = get_mock("test_search.post.response.json")
+
+    app.dependency_overrides[search_dep] = lambda: mock_search
+    yield mock_search
+    del app.dependency_overrides[search_dep]
+
+
+def get_mock(file: str) -> AsyncMock:
```
michal-szostak, line 21: A little more descriptive function name would help a little bit.
Maybe something like `make_mock_from_file` or something similar.

michal-szostak (approved):

Issue #712, opened 2023-07-05
Amendments to Interoperability Guidelines UI


eldenkat:
1. Reference: https://jira.eoscfuture.eu/browse/TECHRDMP-808
 
Data should print on the UI as follows:
Print attribute field name: Related Standard URI
Print content from API:
<relatedStandardURI>https://aarc-community.org/policies/policy-development-kit/
</relatedStandardURI>
Do not print attribute field name 'Related Standard'
Print content from API:
<relatedStandardIdentifier>policy development kit</relatedStandardIdentifier>
Therefore, for each pair transported on the API, the data would appear in the UI as follows:

**Related Standard URI
https://aarc-community.org/policies/policy-development-kit/
policy development kit**

2.  Reference: https://jira.eoscfuture.eu/browse/TECHRDMP-802

- [ ] As a general rule, do not print attribute field names when data is [null]
- [ ] Do not print field name 'Creator Name:'
- [ ] Do not print field name or field data for 'Given Name'
- [ ] Do not print field name or field data for 'Family Name'
- [ ] When Name Identifier = null, do not print either field name or '-'
- [ ] When Affiliation Identifier = null, do not print either field name or '-'
![image-2023-07-04-09-45-11-075](https://github.com/cyfronet-fid/eosc-search-service/assets/113451713/4a2ce971-c08a-4366-890f-2d8e6cad068d)

PR #715, opened 2023-07-07
feat: #712 hide fields Guidelines

Anok666:


tomaszandel (approved):

Marcelinna:
image

Marcelinna:
image

Marcelinna (approved):

Marcelinna (dismissed):
If we dont have value, header should be not visible too
image

Marcelinna:
image
Domain the same situation as above

Marcelinna:
Despite I have value for Name Type*
I don't see the value in SS
image
image

Marcelinna:
I have Name Type: Personal, despite value is organizational
image
image

eldenkat:
Additional changes required

- [ ]  Remove "eosc." from the Provider information -> Provider:  (now: eosc.geant/should be: geant) [**warning**: in case it is possible act alternatively: Required fix is that the Provider name should print as in the Service preview in the MP
image
Alternative:
https://cyfronet-fid.slack.com/files/U041F9A92UV/F05KWRNACAX/zrzut_ekranu_2023-08-02_105733.png
- [ ] Provider Information panel: **Remove 'Provider'+ "providerId": "string"** 
- [ ] Provider Information panel: **Remove 'Catalogue' + Catalogue ID (this is not required)**
- [ ] Provider Information panel: **change title to 'Creator information'** and **flow Creator data over all three columns**
- [ ] Do not print attribute field names when data is null (e.g. currently printing 'EOSC Integration Options = -'

eldenkat:
Overall it looks ok, but the spacing doesn't match:

Zrzut ekranu 2023-08-07 105907

eldenkat:
Domain is empty, should't be displayed.
Zrzut ekranu 2023-08-08 072507

eldenkat (approved):

ui/apps/ui/src/app/pages/guidelines-page/guideline-detail-page.component.ts
```diff
@@ -33,6 +36,11 @@ export class GuidelineDetailPageComponent implements OnInit {
     private _router: Router
   ) {}
 
+  parseCreators(jsonDeserialized: string | undefined): Partial<ICreators>[] {
+    if (jsonDeserialized) console.log(JSON.parse(jsonDeserialized));
```
michal-szostak, line 40: `if`s without brackets can lead to issues - I would add them.

michal-szostak (approved):

ui/apps/ui/src/app/pages/guidelines-page/guideline-detail-page.component.ts
Anok666: true

Anok666: done

michal-szostak (approved):

PR #720, opened 2023-07-10
feat: [#532] Add dynamic fetching of max range

### LINKED ISSUE #532: Take maximum duration filter boundary from the largest duration itself
Duration filter in trainings.

Currently, the maximum boundary of that filter is set as 40h by default. 
It should be set dynamically as the largest duration of the training itself (actual maximum value). 

At the moment the largest duration is:
{
        "title":["Open Science Policies: ORDM, FAIR, and tools"],
        "duration":40000
},
40000s -> 11:06:40

So 11:06:40 should be the set as the maximum duration filter boundary

wujuu:
Closes #532.

Separated into 2 commits: backend and UI for ease of review. I can squash the commits later if needed.

----
@Marcelinna check the "trainings" tab in the "duration" filter for QA

### DIFF
--- a/backend/app/routes/web/search.py
+++ b/backend/app/routes/web/search.py
@@ -9,7 +9,7 @@ from pydantic.typing import Literal
 from requests import Response
 
 from app.routes.web.recommendation import sort_by_relevance
-from app.schemas.web import SearchRequest
+from app.schemas.search_request import SearchRequest
 from app.solr.operations import get, search_dep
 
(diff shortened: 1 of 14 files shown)


Marcelinna:
image
@agpul 
now gives approval, but the field has no unit and I'm afraid that in the future it may be equal to different, uncontrollable value, the question is whether such a form as a slider will then be ok

Marcelinna (approved):

Issue #721, opened 2023-07-11
Refactor settings to use pydantic


michal-szostak:
Currently settings in search service use starlette to use env variables as configuration for the application.
The scope of this issue is to refactor `backend/app/config.py` to be a pydantic class with separate class for base settings class (`GlobalSettings`) and following classes inheriting from it - `DevSettings`, `TestSettings` and `ProductionSettings`. The dev settings should have most of the values have a default value, so that developer setup requires minimal env variable setup.
`ProductionSettings` should still have default variables for backwards compatibility, which will be removed later after ensuring that all relevant values are provided on our production setup.

The new settings classes should be fully backwards compatible with the current setup (same env variable names, etc).

Additionally `config.py` file should be renamed to `settings.py`

Environmental variable `ENVIRONMENT` should govern type of settings loaded. `ENVIRONMENT=production` should be the default. Other options: `dev` and `test`

PR #724, opened 2023-07-12
ref: [#721] Replace Starlette config with Pydantic settings

maria-j-k:


tomaszandel (approved):

backend/app/settings.py
```diff
@@ -0,0 +1,182 @@
...
+def get_settings():
+    """Builds an appropriate config basing on env var ENVIRONMENT value"""
+    environment = Environment(os.environ.get("ENVIRONMENT"))
+    config = SETTINGS_MAP[environment.value]()
+    if config.OIDC_AAI_NEW_API:
+        config.OIDC_ISSUER = f"{config.OIDC_HOST}'/auth/realms/core"
+        config.OIDC_JWKS_ENDPOINT = OIDC_NEW_JWKS_ENDPOINT
+        config.OIDC_USERINFO_ENDPOINT = OIDC_NEW_USERINFO_ENDPOINT
+        config.OIDC_TOKEN_ENDPOINT = OIDC_NEW_TOKEN_ENDPOINT
+        config.OIDC_AUTH_ENDPOINT = OIDC_NEW_AUTH_ENDPOINT
```
michal-szostak, line 107: It would be more elegant to move this code into `GlobalSettings` class itself - this kind of hacking can be done using validators (check `pre=True`)

backend/app/settings.py
```diff
@@ -0,0 +1,182 @@
...
+    EOSC_COMMONS_URL: AnyUrl = "https://s3.cloud.cyfronet.pl/eosc-portal-common/"
+    EOSC_COMMONS_ENV: str = "production"
+    EOSC_EXPLORE_URL: AnyUrl = "https://explore.eosc-portal.eu"
+    RELATED_SERVICES_ENDPOINT: AnyUrl = "https://beta.providers.eosc-portal.eu/api/public/interoperabilityRecord/relatedResources"
+
+    class Config:
+        fields = {
+            "STOMP_USER_ACTIONS_TOPIC": {"env": "ESS_STOMP_USER_ACTION"},
+            "STOMP_CLIENT_NAME": {"env": "ESS_QUEUE_CLIENT_NAME"},
+            "STOMP_SSL": {"env": "ESS_STOMP_SSL"},
```
michal-szostak, line 72: When overriding default names of env variables I'd suggest using declarative approach with `Field`:

```
STOMP_USER_ACTIONS_TOPIC = Field("/topic/user_actions", env="ESS_STOMP_USER_ACTION")
```

This way the name of the variable is in the same place as the variable declaration which might help avoid confusion.

michal-szostak (changes requested):
It is good generally - there are a few things that can be improved to make it a little bit less "hacky". All in all good job.

michal-szostak (approved):

PR #773, opened 2023-08-02
feat: #740 - endpoint to add favourites RP

maria-j-k:


backend/app/routes/web/research_product.py
```diff
@@ -0,0 +1,57 @@
...
+@router.get("/research-product/{resource_type}/{rp_id}")
+async def get_rp_by_id(
+    resource_type: Collection, rp_id: str, solr_get=Depends(get_dep)
+) -> ResearchProductResponse:
+    """
+    Main function responsible for getting details for a given Solr document.
+    args:
+        resource_type: string literal - one of permitted collections
+        rp_id: ID of a research product to be found
+    """
```
michal-szostak, line 28: Docstring here will be visible in swagger - so I'd advice top phrase it more like it would be in API docs. Especially args. If we have literals specyfied here then swagger will correctly show which values are acceptable.

backend/app/routes/web/research_product.py
```diff
@@ -0,0 +1,57 @@
...
+    valid_urls = await asyncio.gather(*[_validate_url(url) for url in urls])
+    return list(filter(None, valid_urls))
+
+
+async def _validate_url(url: AnyUrl) -> Optional[AnyUrl]:
+    """Performs a request to check a link's validity"""
+    response = None
+    async with AsyncClient() as async_client:
+        with suppress(ConnectError):
+            response = await async_client.get(url, follow_redirects=True)
```
michal-szostak, line 54: This call might rise exceptions - https://www.python-httpx.org/exceptions/
We should catch any exception and treat it as an error status.

michal-szostak (changes requested):

backend/app/routes/web/research_product.py
michal-szostak: Network error is not the only exception. But either way - exception here will probably destroy "gather" call.
Additionally server type errors (malformatted https response) is also an error.
If it has been decided (I forgot) that this kind of issue should not be filtered out than we should catch that exception and return url (not `None`).

Current implementation will not return url which triggers error and potentially won't return any response in API.

michal-szostak (approved):

PR #780, opened 2023-08-04
feat: [#606] map oag scientific domain and other pids

### LINKED ISSUE #606: Map scientific domain and other pids
[domains_mapping.zip](https://github.com/cyfronet-fid/eosc-search-service/files/11542995/domains_mapping.zip)
[domains_mapping_poprawione.txt](https://github.com/cyfronet-fid/eosc-search-service/files/11590808/domains_mapping_poprawione.txt)

Acceptance criteria:
- [ ]  As a User I would like to see the same scientific domains in all tabs (the scientific domain for Research products should be the same as for Services)
- [ ] As a user I want to see the scientific domains as a tree (behave as in the Services tab) @aleksandrahojszyk to confirm
- [ ] As a user I would like to see all pids, not only DOI

Michal-Kolomanski:
closes #606 

Before mapping, digits and unnecessary spaces (from the beginning or ending) are removed from OAG scientific domain names.

Assumptions & Rules & Cases:
```
Adjust scientific domains. There is a need to apply the same logic as PC does.
Unfortunately, we cannot enforce anything during onboarding process as they do,
so we need to deal with any case here.
PC's Assumptions:
- scientific domains have 2 levels. Parent -> children depth only,
- you can set only child scientific domain for your resource (parent only is not allowed),
- if you set a scientific domain for your resource, you always add also its parent. Always (child, parent) pairs
- sd children cannot be duplicated,
- sd parents can be duplicated, and they are duplicated a lot!
Abbreviations for rules:
 - number of the same parent strings -> p
 - number of children of the same parent -> ch
Rules how to satisfy PC's assumptions in our cases:
    For each resource we need to check its every sd parent whether PC's assumptions are satisfied.
    Cases:
        1) p == ch -> do nothing (happy path),
        2) p > ch -> delete as many parents to the point where p == ch,
        3) ch > p -> add as many parents to the point where p == ch,
        4) p > 0, ch == 0 -> delete this parent string/strings, but add all his children + parent pairs.
           In other words, when there are only parents without children, add all their (child, parent) pairs,
           but keep p == ch satisfied - so delete those initial parent strings.
 
```


agpul:
Can you enable the filter scientific domain on the front for all catalogs?

tomaszandel (approved):

michal-szostak (approved):

Issue #718, opened 2023-07-10
Single API endpoint for filters


michal-szostak:
Currently to display filters every filter makes a separate request to backend search API in order to retrieve facetes, and possible field values. This is suboptimal as it creates a lot of requests which slows down a browser.
The scope of this issue is to create a separate endpoint in `/api/web/search-filters` which should return a map of all filters (which should be a modified concatenation of the responses from the separate filters requested now.

Currently responses have format:

```
{
    "numFound": 4648283,
    "nextCursorMark": "*",
    "results": [],
    "facets": {
        "count": 4648283,
        "unified_categories": {
            "buckets": [
                {
                    "val": "Discover Research Outputs",
                    "count": 4647889
                },
                ...
```

The response format of the new endpoint should be:

```
{
"<filter name (eg. unified_categories)>": [
                {
                    "val": "Discover Research Outputs",
                    "count": 4647889
                },
                ...
]
}
```

All requested filters should be in the response.

The new endpoint should take all basic parameters as search endpoint and perform all solr queries (preferably concurently).
The response should be cached for the given set of input params. Caching time should be configurable via env variable (and be defined in config). The default should be 5 minutes.

**Acceptance criteria:**

No regressions regarding working of filters
When inspecting network requests number of requests should be reduced to only one`/api/web/search-result` and one`/api/web/search-filters`

agpul:
Estimation: 1day

Issue #603, opened 2023-05-16
Improve "Language" filter


agpul:
Acceptance criteria:

- [ ] As a User I want to see records in the Language filters sorted alphabetically
- [ ] As a User I don't want to see "not specified" record (please remove it)

PR #794, opened 2023-08-08
feat: #718 single api endpoint for filters

Barmin93:
Files related to external API requests (including SOLR has been changed). Please check for potential regressions.


tomaszandel (approved):
looks ok

agpul:
image
no filter visible!

Marcelinna:
Please check this
1. I am on guidelines collection
image
2. Change collection to bundle - I see request to /api/web/search-result and one/api/web/search-filters
image
3. Select on of the filetrs - I see only request to /api/web/search-result - @michal-szostak @Barmin93 it is ok?
image
4. Turn of seletced filter - I see two request 
image
5. Select agian filter from step 3 - I see two request 
image

Marcelinna:
the same situation as above
1. Go to Data collection
2. Click on the tag - I see one request 
image
3. Click another tag, I see two request

Marcelinna:
- [ ] QA
- [ ] DEV
image

Marcelinna:
- [ ] QA
- [ ] DEV
image

Marcelinna:
- [ ] QA
- [ ] DEV
image

Marcelinna:
- [ ] QA
- [ ] DEV
image

Marcelinna:
- [ ] QA
- [ ] DEV
image

Marcelinna:
@michal-szostak When I click on filter for training I see 4 request, I should see 2 ?
image

michal-szostak:
@Marcelinna - in trainings those additional requests come from "Durantion" and "datetime" filters (there are both there). So this is in line with what is to be expected.

Marcelinna:
@michal-szostak we have other behaviour than on production
image
image

Marcelinna:
When I type something in a basic search I see that filters have a different order than in production enviroment


image

Marcelinna:
@michal-szostak 
image

Marcelinna:
@agpul @aleksandrahojszyk @Karolina9824 
when filters do not have any value to select when narrowing down values, the filter header disappears

image

agpul:
> @agpul @aleksandrahojszyk @Karolina9824 when filters do not have any value to select when narrowing down values, the filter header disappears
> 
> image

Yes, this is good behavior. We wanted to do this in the other issue.

michal-szostak:
> When I type something in a basic search I see that filters have a different order than in production environment
> 
> image

This is not changed, but I believe this behaviour is correct - the sorting is done first by `count` (number of documents) and second alphabetically.

The original behavior was a little bit undetermined.

@agpul - I can hack away to make it work the same, but I think that this is clearer...

agpul:
> > When I type something in a basic search I see that filters have a different order than in production environment
> > image
> 
> This is not changed, but I believe this behaviour is correct - the sorting is done first by `count` (number of documents) and second alphabetically.
> 
> The original behavior was a little bit undetermined.
> 
> @agpul - I can hack away to make it work the same, but I think that this is clearer...

We agreed during the daily with Roksana, that this behaviour is OK and can stay.

Marcelinna:
beta 
image
pr
image

We had PR that we wanted language sortable alphabetically
https://github.com/cyfronet-fid/eosc-search-service/issues/603

Marcelinna:
beta - clear button is visible 
image
pr 
image

Marcelinna:
1. Go to Training and check open access filter 
image
2. Change to Services - not all filters are visible
image

Marcelinna (approved):

PR #800, opened 2023-08-13
feat: [#754] Remove bundles from top of search

### LINKED ISSUE #754: Hide bundles from the top of search
Currently, bundles are at the top of search results
image

The scope of the ticket is to: 
- when search phrase and filters are not used - hide bundles from search results.

When bundles will have a dedicated page, they would not be hidden anymore.

wujuu:
Closes #754.

This is a workaround. Fix once bundles are presentable again.

How to test:
1. Go to /search/all
2. Bundles should be at the very bottom of the list (this will be impossible to check manually, unless you go through X amount of pages)
3. Bundles should still be present when you search for them.

### DIFF
--- a/backend/app/routes/web/search.py
+++ b/backend/app/routes/web/search.py
@@ -1,7 +1,10 @@
+# pylint: disable=fixme
+
 """The UI Search endpoint"""
 import itertools
 import logging
 from json import JSONDecodeError
+from typing import Optional
 
 from fastapi import APIRouter, Body, Depends, HTTPException, Query, Request
 from httpx import AsyncClient, TransportError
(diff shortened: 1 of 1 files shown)


Marcelinna:
bundle still high on the list
image

tomaszandel (changes requested):
this sort will not fix it, if we want to be sure, i would recommend based on extend_results_with_bundles , get al bundles from search results and append them to the end when creating putput in create_output function out['results'] take all except bundles and + `    bundle_results = list(
        fillter(lambda doc: doc["type"] == "bundle", res_json["response"]["docs"])
    )`

wujuu:
@tomaszandel I will try your approach but it works with all other datatypes (tested locally) - why would it not work on bundles?

Granted I don't have any bundles in my local dataset so there might be something different about them

wujuu:
@tomaszandel my solution should work now. The problem was not in the sort query but in the `if collection == "all_collection"` condition. We have different prefixes for beta collections so it was not working on PR but was working for me locally.

Marcelinna (approved):

tomaszandel (approved):

Issue #791, opened 2023-08-08
Add "Source" field on the box view for Publications, Data,Software,Other


agpul:
- [ ] As a User I would like to see the Source button on the box view (as a dropdown)
- [ ] As a User I can see different "versions" of the same publication (downloaded from different Sources) hidden under the Source button
- [ ] As a user I can click on the title of the publication and go to the source url

OpanAire reference: https://explore.openaire.eu/search/find/research-outcomes?page=4
Figma: https://www.figma.com/file/pTiscgrb4ie8R57vJmEhyL/Boxy-Redesign?type=design&node-id=7-2879&mode=design&t=rViMb5bYCcQzoqWT-4

agpul:
Estimation: 2 days

PR #822, opened 2023-08-22
feat: #791 add source field for publications

Anok666:


wujuu (approved):

Anok666:
> I think it shouldn't be visible image

Fix

Marcelinna:
When I select one option source button is changing
image

Marcelinna:
Why I don't see source button for DATA
image

Marcelinna:
@agpul @Michal-Kolomanski 
It is ok that I see one source in OpenAir and two in MP? 
When I click on the Source button I see that the item has DOI, but we don't display it on the resource card, because the DOI field in the data dump is empty?
image

wiktorflorian:
> We show everything we get. From my perspective it is okay. @wiktorflorian could you confirm?

From what I've seen, they do it in OpenAIRE so that if an instance/version of a research product has more than 1 url, they just take 1 from the top. Even if they point to different locations. I can check what the situation is in this exact case though, as if there was a link to the research product somewhere it would go faster for me.

Issue #786, opened 2023-08-07
add Show more functionality if there are too many tags


agpul:
Acceptance criteria:
- [ ] As a User  I would like to see tags under the description 
- [ ] As a User I would like to see "+ x more" functionality if there are too many tags (only 3 lines of tags are acceptable) 
![image](https://github.com/cyfronet-fid/eosc-search-service/assets/43000481/74d7fe68-ad3e-4b68-b225-8de3d122b160)

Figma:
https://www.figma.com/file/pTiscgrb4ie8R57vJmEhyL/Boxy-Redesign?type=design&node-id=7-2879&mode=design&t=3kJavuus7ieNTgsJ-4

PR #830, opened 2023-08-23
feat: #786 move tags under description

Anok666:


Marcelinna:
empty space added
image

Marcelinna:
I don't see keywords
image

PR #850, opened 2023-09-01
Before transform service - DO NOT MERGE ME

Michal-Kolomanski:
My local branch for batch data upload (OAG).
DO NOT MERGE ME


Michal-Kolomanski:
Only branch is needed

PR #855, opened 2023-09-04
fix: [#266] Fix duplicated filters

### LINKED ISSUE #266: Tags duplicate filter values on click
image

Step to reproduce

1. Go to [Trainings]
2. Click on License for one of the trainings result
3. Click second time to License 
4. See problem - Filter is duplicated
image

The same case for language in Services
image

wujuu:
Closes #266.

There was a typo when checking for duplicates in tags.


Michal-Kolomanski (approved):
Tested, it works. However, I would reconsider how exactly the behaviour should look like when a user clicks once again on the same filter value @agpul @aleksandrahojszyk

Marcelinna (approved):

PR #864, opened 2023-09-06
732 add the feedback mechanism on every page in search service

Barmin93:


Marcelinna:
I don't see feedback for on details page
image

Barmin93:
> I don't see feedback for on details page image

Added

Michal-Kolomanski (approved):
LGTM :clap: 
I would only consider dynamically changing the background color (from black to white) and the color of a question mark (from white to black) of the feedback widget - when the feedback widget is on the footer. @agpul @aleksandrahojszyk 
image

michal-szostak:
I think this is the PR which can be closed as #892 is the latest changes.

PR #867, opened 2023-09-07
feat: #786 move tags under description

Anok666:


Marcelinna:
Please change the commit description - this PR is about show button for keywords

Marcelinna:
Which rules do you use for the "show more" button for keywords?
image

Marcelinna:
I see the problem with pagination
I don't see the problem on the beta instance
image
image

ui/apps/ui/src/app/layouts/secondary-tags/secondary-tags.component.ts
```diff
@@ -78,6 +84,48 @@ export class SecondaryTagsComponent implements OnChanges {
...
+  temp: IValueWithLabel[] = [];
+
+  constructor(private cdr: ChangeDetectorRef) {}
+
+  ngAfterViewInit() {
+    this.cdr.detectChanges();
+  }
+
+  ngAfterContentChecked(): void {
+    this.cdr.detectChanges();
```
tomaszandel, line 98: smart but.. it wont call onChanges, onChanges called only when inputparams changed, even if you call detectChanges still onChanges not called, i would still implement onChanges interface.. Please double check if what you did here is intentional or a mistake ;)

Anok666: I made a mistake. Old unused code

Anok666:
> URL

I cant repeat this behavior / bug.

PR #871, opened 2023-09-08
857 fix prefix behaviour

maria-j-k:
1. Front-end should not interfere with the prefix. Prefix should be set in the functions interacting directly with Solr, basing on environmental variable COLLECTIONS_PREFIX
2. Solr search functions (search and advance search) should not return Response object, but object consisting of collection name and it's results.
3. Error handling for solr functions has been added


backend/app/consts.py
```diff
@@ -0,0 +1,48 @@
...
+    "data_sources",
+    "datasets",
+    "other_research_product",
+    "publications",
+    "services",
+    "trainings",
+    "software",
+]
+
+COLLECTION_TO_PANEL_ID_MAP = {
```
Michal-Kolomanski, line 34: Please add a comment "Mapping solr collection name -> recommender collection name (panel_id)" or sth similar. 
Consider placing `COLLECTION_TO_PANEL_ID_MAP` above `PanelId`

backend/app/consts.py
```diff
@@ -0,0 +1,48 @@
...
+    "data_source": "data-sources",
+    "dataset": "datasets",
+    "other_rp": "other_research_product",
+    "publication": "publications",
+    "service": "services",
+    "training": "trainings",
+    "software": "software",
+}
+
+PROVIDER_QF = "title^100 description^10 scientific_domains^10"
```
Michal-Kolomanski, line 45: I would also consider creating two separate files. One for collection names, second for solr-specific stuff.

backend/app/consts.py
```diff
@@ -0,0 +1,48 @@
...
+
+# There must be a unique field in the sort for the cursor paging to work.
+DEFAULT_SORT = ["score desc", "id asc"]
```
Michal-Kolomanski, line 48: Please also move different sorts here, from https://github.com/cyfronet-fid/eosc-search-service/blob/01cf2bb0634e83201cca0b3c6ef53574fb15718a/backend/app/routes/web/search.py#L227

Example:
`DMR_SORT = ["publication_date desc"]`

backend/app/recommender/router_utils/common.py
```diff
@@ -1,24 +1,12 @@
...
-    "dataset",
-    "software",
-    "other",
-    "training",
-    "service",
-    "data-source",
-    "bundle",
-    "guideline",
-]
 RE_INT = re.compile("^[0-9]+$")
```
Michal-Kolomanski, line 10: Is it used anywhere?

backend/app/recommender/router_utils/recommendations.py
```diff
@@ -17,19 +17,19 @@
 from app.settings import settings
 from app.solr.operations import get, search
 
-ALL_COLLECTION = f"{settings.NG_COLLECTIONS_PREFIX}all_collection"
+ALL_COLLECTION: Collection = "all_collection"
```
Michal-Kolomanski, line 20: It should be moved to consts.py

backend/app/recommender/router_utils/recommendations.py
```diff
@@ -78,27 +77,28 @@ async def get_recommended_items(client: AsyncClient, uuids: list[str]):
...
-            q="*",
-            qf="id",
-            fq=fq,
-            sort=["id desc"],
-            rows=rows,
-        )
-    if response.status_code != HTTP_200_OK:
+    if collection == "data_source":
+        collection = "data source"
+    if collection == "all_collection":
```
Michal-Kolomanski, line 85: "all_collection" -> ALL_COLLECTION

backend/app/recommender/router_utils/recommendations.py
```diff
@@ -78,27 +77,28 @@ async def get_recommended_items(client: AsyncClient, uuids: list[str]):
...
-            client,
-            ALL_COLLECTION,
-            q="*",
-            qf="id",
-            fq=fq,
-            sort=["id desc"],
-            rows=rows,
-        )
-    if response.status_code != HTTP_200_OK:
+    if collection == "data_source":
```
Michal-Kolomanski, line 83: I think consts.py should have variables for each collection name. So here DATA_SOURCE="data_source" in consts.py and import.

backend/app/recommender/router_utils/recommendations.py
```diff
@@ -78,27 +77,28 @@ async def get_recommended_items(client: AsyncClient, uuids: list[str]):
...
+        collection = "data source"
```
Michal-Kolomanski, line 84: you can rename this `collection` into `type` for example, because it is a type

Michal-Kolomanski (commented):
Good job! Minor changes requested. Also, deep testing would be necessary here to see whether we don't have any regression

backend/app/recommender/router_utils/recommendations.py
maria-j-k: I would have to pick another variable name (type is reserved) and introduce this variable for all other collections. Maybe it could stay as is? In the `fq` field it's kind of 'renamed' to 'type'

Michal-Kolomanski:
Rebase, squash, resolve conversations that have been handled

Marcelinna (approved):

Issue #888, opened 2023-09-18
Problem with long title


Marcelinna:
it does not fit in box
![image](https://github.com/cyfronet-fid/eosc-search-service/assets/59872096/66be49e3-0ba3-4b89-9e59-122ae1434270)

PR #893, opened 2023-09-18
fix for long names without space

jarekzet:
Added fix for long result titles without any space in one line:

image

Closes: #888

### DIFF
--- a/ui/apps/ui/src/styles.scss
+++ b/ui/apps/ui/src/styles.scss
@@ -575,9 +575,9 @@ h6 {
 }
 
 .results #container {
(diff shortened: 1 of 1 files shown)


Marcelinna:
image

Marcelinna:
image

tomaszandel (approved):

tomaszandel:
fix lint also before commit, if it fails it means we have bad lint commited somewhere, so just run lint fix to correct the file if possible and recommit, i think that just word-wrap for <b> is not enough..

Marcelinna (approved):

PR #897, opened 2023-09-19
fix: revert bundle url

michal-szostak:
* Add pin to... to services, data-sources and bundles


agpul (approved):

Michal-Kolomanski (approved):
LGTM. Fix UI styles

PR #898, opened 2023-09-19
feat: 775 map provider in training and IG

maria-j-k:
Filters are not in scope of this task, will be fixed in a follow-up ticket


Marcelinna:
I don't see mapping for training 
image

Marcelinna:
@agpul @Marcelinna we want to display the provider abbreviation or name ?

image

Marcelinna:
in this pr, we return the name for the resource provider to?
image

backend/app/solr/utils.py
```diff
@@ -0,0 +1,80 @@
...
+async def map_providers(client: AsyncClient, docs: List) -> None:
+    """Main function responsible for getting provider's human readible name.
+    Calls for every document of the response payload an appropriate function
+    (basing on resource type)
+    """
+    for doc in docs:
+        resource_type = doc["type"]
+        if resource_type not in ["interoperability guideline", "training"]:
+            return
+        await TYPE_TO_FUNCTION_MAP[resource_type](client, doc)
```
michal-szostak, line 80: This is acceptable way of doing this, but I would suggest a little bit different approach (but it's good as it is).

1. `TYPE_TO_FUNCTION_MAP` should have a type - this is especially useful when having callables.
2. It would be even better to have a function wrapping this functionality, which can provide default:

```
def _noop_mapper(client: AsyncClient, doc: Dict) -> None:
    return None

def get_mapper_function(resource_type: str) -> Callable[[AsyncClient, dict[str, Any], None]:
    return TYPE_TO_FUNCTION_MAP.get(resource_type, _noop_mapper)
```

backend/app/solr/utils.py
```diff
@@ -0,0 +1,80 @@
...
+        return None
+
+
+TYPE_TO_FUNCTION_MAP = {
+    "interoperability guideline": _get_provider_data_guideline,
+    "training": _get_provider_data_training,
+}
+
+
+async def map_providers(client: AsyncClient, docs: List) -> None:
```
michal-szostak, line 71: Also, from python 3.10 (or even 3.9) it's recommended to use `dict, list...` instead of `Dict, List, ....`

michal-szostak (approved):

Marcelinna:
Can you add space ?
image

Marcelinna:
pr - lack of provider by 
image
beta 
image

maria-j-k:
> pr - please check resource organisation image beta image
> 
> image

`EUDAT` is not a proper provider ID so Solr request to providers collection returns no results.

Marcelinna:
Problem with filter
image
image

michal-szostak:
I discussed with @maria-j-k the scope and the decision is as follows: Clicking on the provider's name in results should add filter but the text displayed in the active filter will be provider's id (not the name). This is due to the current limitations and will be fixed later, after filters are merged. Therefore clicking on the provider's name should filter (situation with persistent 0 results is not acceptable) but the text will be different.

@agpul and @aleksandrahojszyk - please double check whether this is acceptable. It might confuse the user, but as a temporary case I would let it pass.

Marcelinna (approved):

PR #901, opened 2023-09-20
fix: added 'exact', fixed bundles

maria-j-k:


backend/app/consts.py
```diff
@@ -62,6 +63,7 @@ class PanelId(str, Enum):
     Collection.SERVICE: PanelId.SERVICES,
     Collection.TRAINING: PanelId.TRAININGS,
     Collection.SOFTWARE: PanelId.SOFTWARE,
+    Collection.BUNDLE: PanelId.BUNDLE,
```
Michal-Kolomanski, line 66: It is not the way. We should not send recommendation requests at all for bundles (like for providers or guidelines for example)

Michal-Kolomanski: Not relevant for now

Michal-Kolomanski (approved):

Marcelinna:
@maria-j-k @Michal-Kolomanski I don't see recommendations for training, it is a bug?

Marcelinna:
lack of Pin the Marketplace project button for Datasurce
image

michal-szostak (approved):

Marcelinna (approved):

eldenkat (approved):

PR #909, opened 2023-09-24
feat: [#766] Add download as csv feature

### LINKED ISSUE #766: Add download results functionality
Acceptance criteria:
- [x] As a User I would like to have the possibility to download the first 2000 search results from SS 
- [x] As a User I would like to see the tooltip with the hint "Download the first 2000 results)
- [x] As a user I would like to have the possibility to see the result in my disc in .csv format

Figma:
https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/Marketplace-EOSC?type=design&node-id=4925%3A12580&mode=design&t=GeVzxxPvfmPJS0mm-1

Reference (the behavior should be similar to in OpenAire):
https://explore.openaire.eu/search/find/projects

wujuu:
Closes #766.

For now, I added basic info for each collection ( "title" and "description" for most collections)


Marcelinna:
DATA: title | description | languages | scientific_domains | authors | identifier
PUBLICATION: title | description | documentType | languages | urls | authors | scientific_domains | identifier

what fields should be displayed? and why aren't they the same for all collections?

Marcelinna:
providers
image

ui/apps/ui/src/app/components/download-results/download-results-button.component.ts
```diff
@@ -0,0 +1,130 @@
...
+        src="assets/download-button.svg"
+        alt="download-icon"
+      />
+      Download Results
+    </div>
+
+    <ess-download-results-modal
+      [loading]="modalLoading"
+      [error]="modalError"
+    ></ess-download-results-modal>
```
michal-szostak, line 41: I'm not the biggest fan of how modals were handled here.
It would be better to have modal component which would be passed directly to `this.modalRef = this.modalService.open(DownloadResultsModalComponent)`, and you could still manipulate inputs via `this.modalRef. componentInstance.loading = false`.

But it would be more readable. Creating component which does not have real html template, only `ng-template` which is being passed to the modal is a little bit working around.

michal-szostak (approved):
Generally it is 'alrighty',
But I have 2 issues - first the way how modals were handled (described in the comments to the code). But the second issue is a little bit more fundamental - I do not think that it is a good call to increase querying rows from 100 to 2000 and using the same API for csv AND search.

It would have been more prudent to create new API endpoint, which would return CSV and which could be (for example) throttled differently than search (for example we could limit it to 1 request per 15 seconds, or something reasonable). The way we do it all in the frontend works, but creates event more strange dependencies between components and services.

If this feature needs to be merged I would say it's good enough, but if we have a little bit more time I would do some redesigning.

agpul:
> DATA: title | description | languages | scientific_domains | authors | identifier PUBLICATION: title | description | documentType | languages | urls | authors | scientific_domains | identifier
> 
> what fields should be displayed? and why aren't they the same for all collections from OpenAir?
> 
> @agpul indicate what fields should be in the csv for a specific collection

We've agreed that this fields are OK.

agpul:
Can we extend this feature to use advanced search functionality?

wujuu:
@michal-szostak Thanks for the comments - added separate backend route for downloading results. We're returning a csv in a StreamingResponse now. Also adjusted the modal code on the frontend. It's all in the same file.

@agpul Extended the feature to include advanced search also. The downloaded fields are now hardcoded for each collection ( "title" and "description" for most collections) - Let me know if we wanna return different fields.

backend/app/routes/web/search.py
```diff
@@ -36,7 +39,7 @@ async def search_post(
     sort: list[str] = Query(
         [], description="Solr sort", example=["description asc", "name desc"]
     ),
-    rows: int = Query(10, description="Row count", gte=3, le=100),
+    rows: int = Query(10, description="Row count", gte=3, le=2000),
```
michal-szostak, line 42: Is this still necessary? As we have dedicated download endpoint?

ui/apps/ui/src/app/components/download-results/download-results-button.component.ts
```diff
@@ -0,0 +1,151 @@
...
+            searchMetadata = {
+              ...constructAdvancedSearchMetadata(metadata, routerParams),
+              ...metadata.downloadParams,
+              rows: DEFAULT_NO_OF_RESULTS,
+            };
+          }
+
+          return this._fetchDataService
+            .downloadResults$(searchMetadata, metadata.facets)
+            .pipe(untilDestroyed(this));
```
michal-szostak, line 124: This is only single shot observable (similar to promise) `untilDestroyed` is not necessary here

michal-szostak (approved):
There are a few minor things. Also - rebase is necessary so QA would need to be repeated to check for regressions.

When I suggested more backend style approach I was thinking about getting 100% rid of saving file via JS and just be able to call `.../download-results?...` which should return file which browser would save on disk. But that approach would require redirecting to the error page on some errors, etc. The way it is done currently is also acceptable.

Marcelinna:
image

Marcelinna:
for guidelines provider name is empty
image
for training providers and resource organisation are empty
image

Marcelinna:
for bundle, I think the main offer ID is not necessary 
image
for bundle I think in offer column is to many information
image

Marcelinna:
@agpul @wujuu 
I think for all we should have value, because now sometimes have too many information or empty value


title | type | description |  access type

tomaszandel (approved):
Do przetestowania adv search przy tych zmianach

Anok666 (approved):

Marcelinna:
Step to reproduce 
1. Go to Bundles
2. Go to Trening - see problem page 1 is automatically changed from page 1 to page 2

wujuu:
> Step to reproduce
> 
> 1. Go to Bundles
> 2. Go to Trainings - see problem page 1 is automatically changed from page 1 to page 2

resolved now

eldenkat (approved):

PR #910, opened 2023-09-24
feat: [#847] Adjust bundle result to latest design

### LINKED ISSUE #847: Adjust bundle box to the new design
- [ ] As a user I would like to see bundles filled with fields from Figma: https://www.figma.com/file/pTiscgrb4ie8R57vJmEhyL/Boxy-Redesign?type=design&node-id=318-843&mode=design&t=jliPFFlbr66Wszwp-4

wujuu:
Closes #847 - depending on the follow-up discussion.

What was changed:
- order of the metadata in the bundle box
- correct borders
- new tags

What was not changed and why:
- no secondary tags (keywords), because it is not clear which ones to take. Bundles do not have tags  or keywords themselves. I tried taking the concated offer/service tag_list but then the filtering does not work, because bundles cannot be filtered by tags which they do not have.
- no icons for services/offers - no real way to fetch them in the current schema. We would have to adjust backend to do it (I would create a ticket for that if it is decided that it's needed.
- no information about offer (like RAM) because it is not included anywhere in the backend response (again, we'd have to add this if it is really needed)
- there is no info about best_access_right for bundles as well. We could include the ones from the offers but that's up to the discussion again.


michal-szostak (approved):

aleksandrahojszyk:
Screenshot 2023-09-26 at 09 58 28
2 zmiany w ui: rozkraczona ikona i za duży odstęp od "pin to the mp" a dołem boxa

Marcelinna:
Please remove Research step from box

agpul:
@jarekzet and from Ola: "Bundle i boxy:
podmienić zepsutą ikonę  bundla na dobrą
zmiana koloru z boku na B3A7FE"

jarekzet:
added mods due to comments above:

image

Marcelinna (approved):

tomaszandel (approved):

Issue #947, opened 2023-10-12
Adding Data source as a field to Solr


wiktorflorian:
This ticket requests the addition of a new data source filter.

![Screenshot 2023-10-12 at 11 20 50](https://github.com/cyfronet-fid/eosc-search-service/assets/127978878/7d23ac5b-d304-48c6-8e52-1148c12c3e81)

As part of this task, we will retrieve the data source ID from the tastes OpenAIRE dump and integrate it into the Solr schema. Additionally, we need to verify the existence of the data source ID in the EOSC Marketplace.

- [x] Load data and check for proper addition.
- [x] Verify that each research product has at least one existing ID in the EOSC Marketplace.
- [x] Add a field named data_source, which is an array of strings, to the Solr schema.
- [x] Construct a function to harvest this data and implement it into the existing transform service.

wiktorflorian:
I have retrieved data from the Marketplace API, where we have 71 onboarded datasources. I obtained data from a dump hosted on Google Drive by Miria on September 26th. A datasource field named eoscDsId was added at the instance level. It is a list of strings containing only ids. September's data dump has 32 distinct eoscDsIds, four of which are absent from Marketplace API pids. These four id are found in 3,607 out of 5,493,768 records, with only 2,739 of the records lacking matching ids.

PR #969, opened 2023-10-23
feat: [#947] data source feature (merge into before_transform)

wiktorflorian:
Pull request created as a response to ticket #947.


transform/connectors/mp_pc.py
```diff
@@ -0,0 +1,76 @@
...
+    except requests.RequestException as err:
+        data = None
+        logger.error(
+            f"{data_type} - retrieving data from {data_address} failed. An error occurred: {err}"
+        )
+
+    return data
+
+
+DATA_SOURCE_ADDRESS = os.environ.get("MP_API_ADDRESS") + "/api/v1/ess/datasources"
```
Michal-Kolomanski, line 55: it can be retrieved from env_vars right? `env_vars[ALL_COLLECTION][DATA_SOURCE][ADDRESS]` I think? Then you dont need to load env vars here

wiktorflorian: This was the only way to do it without changing loader.py. But I managed to add to loader.py the same properties that exist in the development branch. Now it gets them from env_vars.

Michal-Kolomanski (approved):
LGTM! :muscle: 
Did you test the solution in terms of resetting the data source list of pids after an update?

wiktorflorian:
Thank you 💪! Yes, I have tested it.

PR #1000, opened 2023-11-07
fix: add scroll functionality to languages

maria-j-k:


Michal-Kolomanski (approved):
:beer: :four_leaf_clover:
I tested but @Marcelinna you can look also ;p

Marcelinna (approved):

PR #1016, opened 2023-11-13
fix: test PR

tomaszandel:
testowy

### DIFF
--- a/ui/apps/ui/src/app/components/filters/filters.service.ts
+++ b/ui/apps/ui/src/app/components/filters/filters.service.ts
@@ -8,26 +8,77 @@ import {
   ITermsFacetResponse,
 } from '@collections/repositories/types';
 import { SearchMetadataRepository } from '@collections/repositories/search-metadata.repository';
-import { Observable, map } from 'rxjs';
+import { Observable, mergeMap } from 'rxjs';
 import { facetToFlatNodes } from '@components/filters/utils';
+import {
+  removeFilterValue,
+  serializeAll,
+} from '@collections/filters-serializers/filters-serializers.utils';
+import { FiltersConfigsRepository } from '@collections/repositories/filters-configs.repository';
+import { ActivatedRoute, Router } from '@angular/router';
+import { IFqMap } from '@collections/services/custom-route.type';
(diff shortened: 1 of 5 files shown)


ui/apps/ui/src/app/components/filters/filters.service.ts
```diff
@@ -8,26 +8,77 @@ import {
...
   ) {}
 
-  fetchTreeNodes$(
+  convertFacets(facets: OriginalStructure): ConvertedStructure {
+    const result: ConvertedStructure = {};
+
+    for (const key in facets) {
+      const facet = facets[key];
+      if (typeof facet !== 'number' && 'buckets' in facet) {
+        // Zakładamy, że facet.buckets.map zwraca (string | number)[]
```
Michal-Kolomanski, line 45: eng

ui/apps/ui/src/app/components/filters/filters.service.ts
```diff
@@ -8,26 +8,77 @@ import {
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        result[key] = facet.buckets.map((bucket: { val: any }) => bucket.val);
+      }
+    }
+
+    return result;
+  }
+
+  /*fetchTreeNodes$(
```
Michal-Kolomanski, line 54: Seems like a dev stuff

Michal-Kolomanski (approved):
From a code perspective it seems like hacking a little bit. The flow got more complicated. However, the most important here are tests. Do we have some examples that on beta don't work, but here they do? @Marcelinna Please test carefully and check whether there are no regressions

ui/apps/ui/src/app/components/filters/filters.service.ts
tomaszandel: yup, but i preffer to keep it there at least for a sec

ui/apps/ui/src/app/components/filters/filters.service.ts
tomaszandel: yup, will refactor after tests

PR #1053, opened 2023-11-22
feat: [#1051] map new sd oag, merge: before trans

Michal-Kolomanski:


tomaszandel (approved):

wiktorflorian (approved):
Really good job with that mapping!

Issue #1014, opened 2023-11-10
Add placeholders for articles in Search Service


agpul:
Acceptance criteria:
- [ ] As a User I would like to see boxes on the right-side panel that link to articles in Knowledge-hub

Each box should have:
- [ ] title
- [ ] tagline
- [ ] arrow (with link)

Figma:
https://www.figma.com/file/wFzD4qH6SwpfCybte3pLlb/EOSC-DRAFTS?type=design&node-id=1800%3A13309&mode=design&t=0oObWhtrPL1wC6bj-1

PR #1061, opened 2023-11-27
feat: [#1014] load logo icon for bundle from MP assets

Anok666:


ui/apps/ui/src/app/components/results-with-pagination/result.component.ts
```diff
@@ -318,6 +320,69 @@ export class ResultComponent implements OnInit {
...
+    } else {
+      img.onload = () => {
+        return true;
+      };
+      return false;
+    }
+  }
+
+  /*
+ *   setLogoUrl(pid: string, type: string): string {
```
wujuu, line 357: can we remove the commented code

wujuu (approved):

ui/apps/ui/src/app/components/results-with-pagination/result.component.ts
Anok666: ofcourse :), done

Issue #1046, opened 2023-11-21
Increase the number of resources that are passed to sort by relevance


Michal-Kolomanski:
Acceptance criteria:

- [ ] As a User I would like to see 250 resources sorted by relevance

Try to increase the cursor itself. Test the solution and response times.
@Marcelinna, do I remember correctly, that there was a bug that prevented from selecting the sort by relevance for some reason? If yes, fix it also in this ticket. 
https://github.com/cyfronet-fid/eosc-search-service/issues/1045

Issue #1045, opened 2023-11-21
Relevance filter problem


Marcelinna:
Step to reproduce
1. Go to Data
2. Select Sort by Date - Most recent
![image](https://github.com/cyfronet-fid/eosc-search-service/assets/59872096/93e50609-acfb-4230-8d89-b07b54874858)
3. Select Sort by Relevance - User see 50 result, but filter label is still Date - Most recent
![image](https://github.com/cyfronet-fid/eosc-search-service/assets/59872096/356ee1f4-4c88-4427-8725-6ba3a4567410)

PR #1066, opened 2023-11-28
feat: [#1046] Increase pagination results for relevance sort

wujuu:
Fix by just extending the pagination results to 250.

That means that the our cursor is now able to fetch 250 results, and then that we can browse through 25 pages until we fetch new results again.

@Marcelinna to test for regression in performance


Michal-Kolomanski:
Notes: 
1) We switched from 50 resources cursor (5 pages) - to 250 resources cursor (25 pages) (everywhere). One cursor is passed to sort by relevance. Are response times satisfactory? @agpul @Marcelinna 
2) Transformer on beta does not have exactly the same data iteration as we do - so some of resources that are passed may not exist for it - so they are not returned. 
3) Before we pass resources to sort by relevance, we sort them by popualrity - and then pass them to sort by relevance. So you can think about sort by relevance as sorting the most popular resources in certain search context by relevance. Do we want to sort by popularity before passing resources to sort by relevance?

Michal-Kolomanski (commented):
From the code side its okay, but the solution influences the whole pagination process - needs testing whether response times are satisfactory and there are no regressions for example in advance search. 
Furthermore, this PR is so short that I would also fix the bug with sorting in general. [Sort by relevance sorting](https://github.com/cyfronet-fid/eosc-search-service/issues/1045)

agpul:
image
I see no result when I use sort by relevance.

Marcelinna:
in some tabs, relevance should be hidden, e.g. all resources or bundle

Michal-Kolomanski:
Also please make an env variable that would enable setting the `MAX_COLLECTION_RESULTS` variable dynamically

wujuu:
@Michal-Kolomanski updated the code with the env var
@Marcelinna fixed the visibility of the sort by relevance in certain collections

Michal-Kolomanski:
All PRs work with the beta facade. 
`Services -> sort by relevance -> returns no results.`
In general, there is a problem with getting recommendations from Marketplace RS through the facade (even for the recommendation panel). Perhaps that is the same issue here. PCSS said that they will release a new version of their Facade today or tomorrow to solve that issue. I will confirm that with them

ui/apps/ui/src/app/collections/data/all/adapter.data.ts
```diff
@@ -165,9 +165,6 @@ export const allCollectionsAdapter: IAdapter = {
       id: string;
     }
   ): IResult => ({
-    isSortCollectionScopeOff: true,
```
Michal-Kolomanski, line 168: I am trying to recall what was the purpose of this variable. Perhaps to completely turn off sort by functionality for e.g. providers? Nonetheless, good job for deleting a deprecated variable 👏

Michal-Kolomanski (approved):
Sort by relevance for services does not work, but it almost certain that the cause is not this PR. In the worst scenario, we can turn off sort by relevance in services collection - its not a blocking issue in terms of going with sort by relevance into production.

ui/apps/ui/src/app/collections/data/all/adapter.data.ts
wujuu: thanks - it just wasn't used anywhere so no point in keeping it

Marcelinna:
@agpul @wujuu @aleksandrahojszyk 
It looks a bit like a bug, it is ok for us?
or maybe we want to add information that if there is relevance, it will always be 250
image

Marcelinna:
In beta we dont display random recomnedation for datasource
beta
image
pr
image

Issue #1067, opened 2023-11-29
Add link to Knowledge hub in the training section


agpul:
Acceptance criteria:
- [ ] As a User I would like to see the link to knowledge on in the training section according to Figma: https://www.figma.com/file/D2vgx66m1ADdBY0A46NdZ6/EOSC-Design?type=design&node-id=6516-9363&mode=design

PR #1069, opened 2023-11-29
feat: [#1067] url box redirect to KH website

Anok666:


aleksandrahojszyk:
approve

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.html
```diff
@@ -89,6 +89,13 @@ <h5>No results</h5>
         "
       ></ess-sort-by-functionality>
     </div>
+   <ng-container  *ngIf="this.type === 'training'">
+    <a href="https://knowledge-hub.eosc-portal.eu">
```
maria-j-k, line 93: Shouldn't we take the url from the config?

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.html
```diff
@@ -89,6 +89,13 @@ <h5>No results</h5>
         "
       ></ess-sort-by-functionality>
     </div>
+   <ng-container  *ngIf="this.type === 'training'">
```
maria-j-k, line 92: Why this condition is based on the type of the first result? 
If all_collection would have a training item as the first result the banner should appear?
We can get the collection from the `CustomRoute`

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.html
Anok666: true, done.

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.html
Anok666: this box with url is only for the training collection. However, I based it on CustomRoute and collections after your comment

Marcelinna (approved):

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
```diff
@@ -42,6 +43,8 @@ export class ResultsWithPaginationComponent implements OnInit {
       this.isError = false;
     }
 
+    this.type = response.results[0].type.value;
```
maria-j-k, line 46: is this still needed?

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
```diff
@@ -30,6 +30,7 @@ export class ResultsWithPaginationComponent implements OnInit {
   } = {};
   isError = false;
   marketplaceUrl = '';
+  type = '';
```
maria-j-k, line 33: is this still needed?

maria-j-k (approved):

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
Anok666: done

ui/apps/ui/src/app/components/results-with-pagination/results-with-pagination.component.ts
Anok666: done

PR #1093, opened 2023-12-06
feat: [#1021] add transformer endpoint for deleting collections

### LINKED ISSUE #1021: Create transformer endpoint for deleting collections
## Purpose of that endpoint:
Delete all collections from a single data iteration for a given solr instance (solr instance is fixed).

1) Check if some aliases are pinned to those collections a user wants to delete. If there is at least one alias pinned to a single collection that user wants to delete - Dont delete any collection and return an appropriate exception. It shouldn't be possible to delete a collection with an alias.

### Endpoint params
`cols_prefix`: str, Optional  By default empty str.
`date`: date, Optional

wiktorflorian:
Pull request created as a response to ticket #1021.

Closes #1021.


Michal-Kolomanski (approved):
LGTM! 💪 👏

PR #1094, opened 2023-12-06
fix: log removed

tomaszandel:


Marcelinna (approved):

Michal-Kolomanski (approved):

backend/app/routes/web/search_results.py
```diff
@@ -87,9 +88,13 @@ async def search_post(
...
             await extend_results_with_bundles(client, res_json)
         if collection in [Collection.ALL_COLLECTION, Collection.GUIDELINE]:
-            res_json["response"]["docs"] = await extend_ig_with_related_services(
-                client, res_json["response"]["docs"]
-            )
+            try:
+                new_docs = await extend_ig_with_related_services(
+                    client, res_json["response"]["docs"]
+                )
+                res_json["response"]["docs"] = copy.deepcopy(new_docs)
```
Michal-Kolomanski, line 95: Is it necessary to create everywhere a deepcopy? We should definitely avoid that as much as possible. That is a costly operation

backend/app/utils/ig_related_services.py
```diff
@@ -51,21 +52,30 @@ async def extend_ig_with_related_services(client: AsyncClient, docs: list[dict])
...
             except RelatedServicesError:
-                return []
-            if related_services_pids:
-                doc["related_services"] = await _get_related_services(
-                    client, related_services_pids
-                )
-            else:
-                doc["related_services"] = []
-    return docs
+                print("related_service_pid error")
```
Michal-Kolomanski, line 64: remember to delete all print statements

PR #1101, opened 2023-12-07
fix: prints removed and logo added

tomaszandel:
- pritns changed to logs
- logo support added

### DIFF
--- a/backend/app/routes/web/search_results.py
+++ b/backend/app/routes/web/search_results.py
@@ -94,7 +94,7 @@ async def search_post(
                 )
                 res_json["response"]["docs"] = copy.deepcopy(new_docs)
             except (Exception,):  # pylint: disable=broad-except
-                print("Something goes wrong..")
+                logger.exception("Exception happened during related services extension")
 
     collection = response.collection
     out = await create_output(request_session, res_json, collection, sort_ui)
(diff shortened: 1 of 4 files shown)


maria-j-k (approved):

backend/app/routes/web/search_results.py
```diff
@@ -94,7 +94,7 @@ async def search_post(
                 )
                 res_json["response"]["docs"] = copy.deepcopy(new_docs)
             except (Exception,):  # pylint: disable=broad-except
-                print("Something goes wrong..")
+                logger.exception("Exception happened during related services extension")
```
Michal-Kolomanski, line 97: `except Exception as e `
Return `e` in the log

Not the biggest fan of such a broad catch. If possible except only concrete exceptions.

ui/apps/ui/src/app/components/results-with-pagination/result.component.html
```diff
@@ -1,6 +1,33 @@
-<div class="type-icon-holder {{ type.label }}">
-  <div class="type-icon"></div>
-</div>
+<ng-container
+  *ngIf="
+    type.value === 'service' ||
+      type.value === 'data-source' ||
+      type.value === 'provider' ||
+      type.value === 'data source';
```
Michal-Kolomanski, line 6: Why `data-source` and `data source`?

Michal-Kolomanski: By the way, offers in bundles also should have icons, right? @agpul

tomaszandel: old code, its just a reformat :)

PR #1110, opened 2023-12-07
fix: [#1108] keywords for mp resources - workaround 

### LINKED ISSUE #1108: Fix keywords presentation in all catalogs

Michal-Kolomanski:
closes #1108


Marcelinna:
why first servies are without keywords ? in all catalogs
image
image

Issue #1099, opened 2023-12-06
Project transformer


wiktorflorian:
Implement a Project transformer based on the input schema. Transform the received project dump into prepared output schema and create Solr schema based on the prepared Solr fields.

### Example dump input:
```
{
    "acronym":"RDA TIGER",
    "callidentifier":"HORIZON-INFRA-2022-EOSC-01",
    "code":"101094406",
    "enddate":"2025-12-31",
    "funding":[
        {"funding_stream":{
            "description":"Horizon Europe Framework Programme - HORIZON Action Grant Budget-Based",
            "id":"EC::HE::HORIZON-AG"
        },
        "jurisdiction":"EU",
        "name":"European Commission",
        "shortName":"EC"
        },
        {"funding_stream":{
            "description":"Horizon Europe Framework Programme - HORIZON Coordination and Support Actions",
            "id":"EC::HE::HORIZON-CSA"
        },
        "jurisdiction":"EU",
        "name":"European Commission",
        "shortName":"EC"
        },
        {"funding_stream":{
            "description":"Horizon Europe Framework Programme - HORIZON-CSA\\HORIZON Action Grant Budget-Based",
            "id":"EC::HE::HORIZON-CSA\\HORIZON-AG"
        },
        "jurisdiction":"EU",
        "name":"European Commission",
        "shortName":"EC"
        }
    ],
    "granted":{"currency":"EUR","fundedamount":3000000.0,"totalcost":3000000.0},
    "h2020programme":[
        {"code":"H2020-EU.1.3.2.","description":"MSCA Mobility"},
        {"code":"H2020-EU.1.3.2.","description":"MSCA Mobility"}
    ],
    "keywords": "porphyrin derivatives,photophysical characterisation,fluorescent molecular markers,fluorescence microscopy, Exact Sciences and Engineering",
    "id":"40|corda_____he::573e591300b1a78859e1881f4d466494",
    "openaccessmandatefordataset":true,
    "openaccessmandateforpublications":true,
    "startdate":"2023-01-01",
    "subject":["Support for initiatives helping to generate global standards, specifications and recommendations for open sharing of FAIR research data, publications and software"],
    "summary":"The RDA TIGER aims to provide services to facilitate and support well-defined Working Groups between key European and international initiatives, resulting in concrete alignment, harmonisation, and standardisation of Open Science developments and technologies globally. 
[...]

PR #1145, opened 2024-01-02
feat: [#1099] Project transformer (merge into before_transform_service)

wiktorflorian:
Pull request created as a response to a ticket #1099


transform/mappings/currency.py
```diff
@@ -0,0 +1,7 @@
+"""Mapping of currency symbols to their corresponding three-letter acronyms."""
+
+currency_mapping = {
```
Michal-Kolomanski, line 3: if we have only 3 currency symbols its fine. If there could be more - I would suggest using some 3party package, 
e.g. https://pypi.org/project/pycountry/ - includes all info about countries. 
OR something more lightweight:
https://pypi.org/project/currency-symbols/

transform/utils/utils.py
```diff
@@ -108,3 +110,17 @@ def extract_digits_and_trim(input_str: str):
     remaining_str = input_str[index:].strip()
 
     return remaining_str
+
+
+def map_currency_symbol(currency_symbol):
+    """Maps a currency symbol to its corresponding value."""
+    return currency_mapping.get(currency_symbol, currency_symbol)
```
Michal-Kolomanski, line 117: its so short that it shouldn't be a function. Just use it directly in the code

transform/utils/utils.py
```diff
@@ -108,3 +110,17 @@ def extract_digits_and_trim(input_str: str):
...
+
+
+def title_case_funct(s):
```
Michal-Kolomanski, line 120: - if it will be used only in projects - please define it as a method in Project class 
- Could you elaborate a little bit more in the doc string on what this function is actually doing? Its not that clear for me :D

transform/transformers/project.py
```diff
@@ -0,0 +1,328 @@
...
+
+        def calculate_eosc_score(
+            dataframe: DataFrame, conditions: List[when]
+        ) -> List[int]:
+            """Calculate the EOSC score and return a list of scores."""
+            eosc_score_column = sum(conditions).alias(EOSC_SCORE)
+            dataframe = dataframe.withColumn(EOSC_SCORE, eosc_score_column)
+            return dataframe.select(EOSC_SCORE).rdd.flatMap(lambda x: x).collect()
+
+        harvested_properties_schema = StructType(
```
Michal-Kolomanski, line 245:
```
        harvested_properties_schema = StructType(
            [
                StructField("currency", StringType(), True),
                StructField("funding_stream_title", ArrayType(StringType()), True),
                StructField("funding_title", ArrayType(StringType()), True),
                StructField("keywords", ArrayType(StringType()), True),
                StructField("total_cost", FloatType(), True),
            ]
        )

        harvested_properties_df = create_df(
            harvested_properties, harvested_properties_schema, spark
        )

        project_df = df.select(
            df[ABBREVIATION],
            df[CODE],
            df[DESCRIPTION],
            df[END_DATE],
            df[OPEN_ACCESS_MANDATE_FOR_DATASET],
            df[OPEN_ACCESS_MANDATE_FOR_PUBLICATIONS],
            df[START_DATE],
            df[SUBJECT],
            df[TITLE],
        )

        joined_df = join_different_dfs((project_df, harvested_properties_df))
```
That should be a part of this method. Its out of its scope and joining Sparks Ddfs is really costly operation. 
Lets have a quick meeting on Monday regarding this issue

transform/utils/utils.py
wiktorflorian: The function has been relocated to the ProjectTransformer class and now includes a warning that verifies if the value is present in the dictionary keys or values.

transform/utils/utils.py
wiktorflorian: deleted

transform/transformers/project.py
wiktorflorian: I added this method as a usability test. If we decide to develop the method, I included a `# TODO REFACTOR`.

transform/mappings/currency.py
wiktorflorian: None of the third-party libraries available, such as forex_python, py-country, or currency-symbols, support the mapping of currency symbols to 3-digit codes. This is because the $ symbol is used by 26 different currencies.

transform/transformers/project.py
```diff
@@ -0,0 +1,326 @@
...
+
+    @staticmethod
+    def map_currency_symbol(self, currency_symbol):
+        """Maps a currency symbol to its corresponding value. Logs a warning if not found."""
+        if (
+            currency_symbol not in currency_mapping
+            and currency_symbol not in currency_mapping.values()
+        ):
+            warning_message = f"Unknown {currency_symbol=}"
+            self.logger.warning(warning_message)
```
Michal-Kolomanski, line 267: you can use it directly: 
`self.logger.warning(f"Unknown {currency_symbol=}")`

Michal-Kolomanski (approved):
LGTM 💪

Issue #1153, opened 2024-01-10
Enhance "Sort by" functionality


eldenkat:
Acceptance criteria:

- [ ] as a user I want to sort by "Recommended" not by "Recommendation"
- [ ] as a user I want to sort by "Downloads & Views" not by "Most popular"
- [ ] as a User I want to sort results in each collection like presented below:

![image](https://github.com/cyfronet-fid/eosc-search-service/assets/113451713/8cb77530-2a93-40e3-af0b-75f10dd581bf)

![image](https://github.com/cyfronet-fid/eosc-search-service/assets/113451713/e4ead7ee-2e76-407e-b8de-1aeb53ff8395)

PR #1173, opened 2024-01-23
fix: sorting names

maria-j-k:
closing #1153


Marcelinna:
@maria-j-k 
image

Can you put Best match higher ?
and delete - from Data - least 
image

Marcelinna (approved):

PR #1199, opened 2024-02-09
feat: add catalogues collection

maria-j-k:
1. Catalogue collection is added as a tab in the top panel
2. redirection to all_catalogues with the chosen catalogue as a filter
3. redirection to mp for the detail view
4. search by abbreviation added also to provider, project and organisation collections

### DIFF
--- a/backend/app/consts.py
+++ b/backend/app/consts.py
@@ -18,6 +18,7 @@ class ResourceType(str, Enum):
     PROVIDER = "provider"
     ORGANISATION = "organisation"
     PROJECT = "project"
+    CATALOGUE = "catalogue"
 
 
 class Collection(str, Enum):
@@ -37,6 +38,7 @@ class Collection(str, Enum):
     OFFER = "offer"
     ORGANISATION = "organisation"
     PROJECT = "project"
+    CATALOGUE = "catalogue"
 
 
 ALL_COLLECTION_LIST = [
@@ -50,7 +52,6 @@ ALL_COLLECTION_LIST = [
     Collection.BUNDLE,
     Collection.OTHER_RP,
     Collection.PROVIDER,
-    Collection.ORGANISATION,
 ]
 
 
@@ -107,6 +108,8 @@ PANEL_ID_OPTIONS = [
 ]
 
 SPECIAL_COLLECTIONS = [Collection.PROJECT, Collection.ORGANISATION]
+
+CATALOGUE_QF = "title^100 abbreviation^100 description^10 keywords_tg^10"
 
 PROVIDER_QF = "title^100 description^10 scientific_domains^10"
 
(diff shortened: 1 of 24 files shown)


Marcelinna:
add logo 
image

backend/app/consts.py
```diff
@@ -131,6 +136,8 @@ class PanelId(str, Enum):
     "r": ["popularity desc"],
     "pdmr": ["start_date desc"],
     "pdlr": ["start_date asc"],
+    "cdmr": ["created_at desc"],
+    "cdlr": ["created_at asc"],
```
Michal-Kolomanski, line 140: Date sorting got out of hand here. 
- here we have 6 args for sorting by date only `"cdmr", "cdlr", "pdmr", "pdlr", "dmr", "dlr"`
- definitely, I would unify `created_at` from catalogs and name it `publication_date` as it is in other data types. @wiktorflorian could you please do that? 
- `pdmr, pdlr` we can even consider renaming "`start_date`" -> `publication_date` if that makes sense

ui/apps/ui/src/app/collections/data/catalogues/adapter.data.ts
```diff
@@ -0,0 +1,51 @@
...
+      {
+        label: 'Legal status',
+        values: toValueWithLabel(toArray(catalogue.legal_status)),
+        filter: 'legal_status',
+      },
+    ],
+    collection: COLLECTION,
+    title: catalogue.title ?? '',
+    abbreviation: catalogue.abbreviation ?? '',
+    description: [catalogue?.description]?.join(' ') || '',
```
Michal-Kolomanski, line 39: @wiktorflorian description is an array in catalogs? Can we safely assume that it is always a string?

Michal-Kolomanski (commented):
👏 only data concerns

backend/app/consts.py
wiktorflorian: I will rename it, on my end.

ui/apps/ui/src/app/collections/data/catalogues/adapter.data.ts
wiktorflorian: As we receive the description as a string from MP, we can assume that it will always be a string.

backend/app/consts.py
Michal-Kolomanski: please ping @maria-j-k when ready

ui/apps/ui/src/app/collections/data/catalogues/adapter.data.ts
Michal-Kolomanski: After consulting, this code is not data-dependent - so no action is needed

backend/app/consts.py
Michal-Kolomanski: @wiktorflorian 
- lets rename only `created_at` -> `publication_date` in catalogues
- lets leave `start_date` as it is in projects & orgs

backend/app/consts.py
```diff
@@ -131,6 +134,8 @@ class PanelId(str, Enum):
     "r": ["popularity desc"],
     "pdmr": ["start_date desc"],
     "pdlr": ["start_date asc"],
+    # "cdmr": ["created_at desc"],
```
Michal-Kolomanski, line 137: just delete :D

Michal-Kolomanski (approved):

Marcelinna (approved):

PR #1252, opened 2024-02-28
fix: [#1212] training scientific domain

### LINKED ISSUE #1212: Map scientific domains for Training Catalogue
Acceptance criteria:
- [ ] As a User I would like to see the scientific domain for Training Catalogue display in proper manner (without _)

Raw input:
image

Scope for now:
- hardcode the mapping as a whole 
- satisfy the needs of current rules:
remember about the rules: https://github.com/cyfronet-fid/eosc-search-service/blob/4ac32f1e613aa1a397230b2ffe6f1aeff2000a66/transform/app/transform/utils/common.py#L175. 
Resources are bound to scientific subdomains but we need also to add parent for every child.

Example:
So e.g. when some resource has a scientific subdomain:
RAW:
`scientific_subdomain-engineering_and_technology-other_engineering_and_technology_sciences`

Expected output from that one scientific subdomain:
`"Engineering & Technology>Other Engineering & Technology Sciences", "Engineering & Technology"`

Scope when PC will repair the bug:
- from a standard get trainings query (https://providers.eosc-portal.eu/api/public/trainingResource/all?catalogue_id=all&active=true&suspended=false&quantity=10000 or beta) get scientific subdomains. PC does not return root scientific domains in public API - reported to them as a bug.
- hardcode root scientific domain mapping
- map scientific domains to the already established form: https://github.com/cyfronet-fid/eosc-search-service/blob/2a44f008ba226b687777f9c32f0559486eed530e/transform/app/transform/mappings/scientific_doamin.py#L5

wiktorflorian:
Closes #1212 

In addition to adding mapping for scientific domains, `black .` was used to format other files.

### DIFF
--- a/transform/app/transform/mappings/scientific_doamin.py
+++ b/transform/app/transform/mappings/scientific_doamin.py
@@ -863,3 +863,27 @@ scientific_domains_mapping = {
     "wireless sensor network/wireless networking": "Natural Sciences>Computer & Information Sciences",
     "zoology": "Agricultural Sciences>Animal & Dairy Sciences",
 }
+
+# TODO: add dynamic mapping when PC will return proper facets from public API
+sd_training_temp_mapping = {
+    "agricultural_sciences": "Agricultural Sciences",
+    "agricultural_sciences-agricultural_biotechnology": "Agricultural Sciences>Agricultural Biotechnology",
+    "engineering_and_technology": "Engineering & Technology",
+    "engineering_and_technology-civil_engineering": "Engineering & Technology>Civil Engineering",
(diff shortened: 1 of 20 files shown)


transform/app/transform/mappings/scientific_doamin.py
```diff
@@ -863,3 +863,26 @@
     "wireless sensor network/wireless networking": "Natural Sciences>Computer & Information Sciences",
     "zoology": "Agricultural Sciences>Animal & Dairy Sciences",
 }
+
+sd_training_temp_mapping = {
```
Michal-Kolomanski, line 867: Add a comment: TODO: add dynamic mapping when PC will return proper facets from public API

Michal-Kolomanski (approved):
👏

Issue #1307, opened 2024-04-15
Finish live update


Michal-Kolomanski:
Scope:
- [x] debug suspend in providers - why event though all tasks are successful, the provider is not deleted from solr?
- [x] if data source is updated we need to check if in our solr service with the same ID exists - if so, delete that service and proceed with data source update task
- [x]  if service is updated we need to check if in our solr data source with the same ID exists - if so, delete that data source and proceed with service update task

wiktorflorian:
In case of providers there were 2 issues. The first issue was that, upon suspending or deactivating a provider, the MP sen four tasks: three with `delete_data_by_id` and one `update`. The `update` task started simultaneously with the others but took longer to execute, resulting in no changes in Solr. The second issue involved incorrect behavior of the function `delete_data_by_id` due to an error in its implementation. It was only deleting records from the first collection, typically `all_collection`. This issue has been corrected with my PR.

![Screenshot 2024-04-24 at 11 19 25](https://github.com/cyfronet-fid/eosc-search-service/assets/127978878/f32d84a1-4328-4998-950b-6a343d245163)

Additionally, my PR includes logic to handle updates to the data source and service

PR #1310, opened 2024-04-24
fix: [#1307] service and data source update

wiktorflorian:
Refers to #1307


transform/app/tasks/delete_data_by_id.py
```diff
@@ -13,16 +13,18 @@
 @celery.task(name="delete_data_by_id")
 def delete_data_by_id(
     col_name: str, data: dict | list[dict], delete: bool = True
-) -> dict | None:
+) -> list:
```
Michal-Kolomanski, line 16: you can add -> list[dict]

transform/app/api/routes/transform/batch.py
```diff
@@ -33,9 +34,20 @@ async def batch_update(
...
-        return {"task_id": task.id}
+        if data_type == "service":
+            task_ids.extend(update_service(data))
+        elif data_type == "data source":
+            task_ids.extend(update_data_source(data))
+        else:
+            task = transform_batch.delay(data_type, data, full_update=False)
+            task_ids.append(task.id)
+
+    if task_ids:
```
Michal-Kolomanski, line 50: just `return {"task_ids": task_ids}` is fine.
data_type & action are Literals so validation is already there

transform/app/api/routes/transform/batch.py
```diff
@@ -33,9 +34,20 @@ async def batch_update(
     data: dict | list[dict],
 ):
     """Transform a batch of a single type of data. Used for live update"""
+    task_ids = []
```
Michal-Kolomanski, line 37: I believe it could be just a dict with key-value pairs. 
There may be more than 1 task, but still it is about a single event

transform/app/transform/live_update/data_type_handlers.py
```diff
@@ -0,0 +1,37 @@
+from app.tasks.batch import transform_batch
+from app.tasks.delete_data_by_id import delete_data_by_id
+from app.services.solr.validate import check_document_exists
+from app.transform.schemas.properties.data import ID
+from app.settings import settings
+
+
+def update_data_item(data, data_type, linked_data_type):
```
Michal-Kolomanski, line 8: -> dict

transform/app/transform/live_update/data_type_handlers.py
```diff
@@ -0,0 +1,37 @@
...
+        data_type (str): The type of the data being updated.
+        linked_data_type (str): The type of the data linked to the main data, which may need deletion.
+
+    Returns:
+        list: A list of Celery task IDs.
+    """
+    tasks = []
+    if check_document_exists(linked_data_type, data[ID]):
+        delete_task = delete_data_by_id.delay(linked_data_type, data, delete=True)
+        tasks.append(delete_task.id)
```
Michal-Kolomanski, line 23: please add those tasks as key-value pairs in a dict - key should tell whether it is delete or update task :D

Michal-Kolomanski (approved):
💪

transform/app/api/routes/transform/batch.py
wiktorflorian: Yes, you are right

transform/app/transform/live_update/data_type_handlers.py
wiktorflorian: Yes, I forgot the type hints...

Michal-Kolomanski (approved):

PR #1320, opened 2024-05-22
feat: [#820] skeleton for dump upload

### LINKED ISSUE #820: Create an endpoint for dump upload

Michal-Kolomanski:
closes #820


transform/app/tasks/transform/dump/process.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return CeleryTaskStatus(status="toBeImplemented").dict()
+
+
+@celery.task(name="transform_dump")
+def process_dump(req_body: dict) -> AsyncResult:
+    """Main celery task for managing of processing a dump of data.
+    Workflow:
+        1) Validate dump
+        2)* Create solr collections
```
wiktorflorian, line 36: Just a typo

Michal-Kolomanski: Actually, it is done on purpose because the second step is optional. Someone may send data only to S3

wiktorflorian: I see, maybe there should be a side note at the end of the docstring? Like: `* - optional step'? If not, leave it as it is.

wiktorflorian (approved):
Really nice 🩻 for the dump upload, waiting for more 👀

wiktorflorian (approved):

transform/app/tasks/transform/dump/process.py
Michal-Kolomanski: Added " - optional "

PR #1342, opened 2024-07-18
feat: [#1341] SE & DB live documentation

### LINKED ISSUE #1341: Decide what should be kept in search engine & DB
Scope:
- [x] Which properties should be kept in the search engine?
- [ ] Come up with search engine schema
- [x] What should be kept in DB?
- [x] configure sphinx
- [ ] configure readthedocs
- [x] From where should we get filters and their counts?

Michal-Kolomanski:
closes #1341


wiktorflorian (approved):
Wow, really amazing work 💪

PR #1348, opened 2024-07-31
feat: add sentry integration

maria-j-k:


Michal-Kolomanski (commented):
I didnt mention this but please integrate Celery and Redis with Sentry also. That will be extremely useful for the transformer which is a part of Search Service.

Example: https://github.com/cyfronet-fid/recommender-system/blob/5db4662b7ed1d4e3909a2686fb56920c422b0f57/recommender/__init__.py#L190

PR #1402, opened 2024-09-10
feat: [#1401] load data from cloud (s3)

### LINKED ISSUE #1401: Load data from a cloud (S3 bucket)
Scope:
- [ ] Create a celery task that will load data from a link to S3 folder
- [ ] The task should return a python object (e.g. dict) where keys identify data type (preferably names that are used in our code - settings.py) and values enable quick and easy access to specific files from that directory. Do not load whole files there. Use generators or not - if you think that those values can be used multiple times. You can store full links to those files inside the directories. Dont story only names because then later on we will be forced to pass URL further in the code
- [ ] It should expect a specific dump structure - e.g. certain directory names. 
- [ ] It shouldn't expect a specific file extension - .json, .zip, ... are all fine 
- [ ] Do not require all directories to be present - e.g. alone folder called "dataset" is just fine - then just return dict with:
```
{
"dataset": <values>
}
```

wiktorflorian:
Closes #1401

### DIFF
--- a/transform/app/tasks/transform/dump/load.py
+++ b/transform/app/tasks/transform/dump/load.py
@@ -1,17 +1,32 @@
 """A celery task for loading a data dump."""
 
 import logging
-from typing import Optional
+from typing import Dict, Optional
 
 from app.services.celery.task import CeleryTaskStatus
+from app.services.celery.task_statuses import FAILURE, SUCCESS
+from app.tasks.utils.s3_paths import get_s3_paths_task
 from app.worker import celery
 
 logger = logging.getLogger(__name__)
 
 
 @celery.task(name="load_dump")
-def load_dump(prev_task_status: Optional[CeleryTaskStatus], s3_url: str) -> dict:
+def load_dump(prev_task_status: Optional[CeleryTaskStatus], s3_url: str) -> Dict:
     """Task to load the dump"""
-    logger.info("Loading dump...")
-    # TODO
-    return CeleryTaskStatus(status="toBeImplemented").dict()
+    if prev_task_status and prev_task_status.get("status") == SUCCESS:
+        logger.info("Loading dump...")
+
+        logger.info("Getting S3 paths...")
+        get_s3_paths_task(s3_url)
+        logger.info("S3 paths retrieved")
+
+        logger.info("Transforming...")
+        # TODO TRANSFORM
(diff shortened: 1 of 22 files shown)


transform/app/tasks/transform/dump/validate.py
```diff
@@ -16,12 +24,31 @@ def validate_dump(prev_task_status: Optional[CeleryTaskStatus], s3_url: str) ->
...
+
+        # 3. Validate folder structure in the dump
+        validate_s3_directory_structure(s3_client, s3_url)
+
+        # 4. Validate file extensions
+        validate_files_extension(s3_client, s3_url)
+
+        logger.info("Dump validated successfully.")
+
+        return CeleryTaskStatus(status="validated").dict()
```
Michal-Kolomanski, line 51: Lets set the status as `success` or `failure` - please set `failure`  when it failed. Each task should assign the same statuses.
Lets make some python file called celery_tasks_statuses and there assign:
SUCCESS =  `success` ...
Later on import those and set inside celery tasks. If you can please refactor other celery tasks to use the same statuses

Michal-Kolomanski: or you can name it successful and unsuccessful - decide on your own :D

transform/app/tasks/transform/dump/load.py
```diff
@@ -1,17 +1,31 @@
...
 
 
 @celery.task(name="load_dump")
-def load_dump(prev_task_status: Optional[CeleryTaskStatus], s3_url: str) -> dict:
+def load_dump(prev_task_status: Optional[CeleryTaskStatus], s3_url: str) -> Dict:
     """Task to load the dump"""
-    logger.info("Loading dump...")
-    # TODO
-    return CeleryTaskStatus(status="toBeImplemented").dict()
+    if prev_task_status and prev_task_status.get("status") == "validated":
```
Michal-Kolomanski, line 16: "validated" will be refactored

transform/app/tasks/transform/dump/validate.py
wiktorflorian: Added new `task_statuses.py` in `app.services.celery`, with 3 states, which are used when `CeleryTaskStatus` is returned

Michal-Kolomanski (approved):
Really good job :D please squash commits

PR #1453, opened 2024-10-21
fix redirect link

maria-j-k:


ui/apps/ui/src/app/components/results-with-pagination/result.component.ts
```diff
@@ -85,6 +85,29 @@ export class ResultComponent implements OnInit {
     return '';
   }
 
+  get redirectUrl(): string | null {
+    console.log('tadam');
```
goreck888, line 89: Probably unnecessary console.log

wiktorflorian (approved):
LGTM

goreck888 (approved):

Issue #1546, opened 2025-04-28
Raplace loading icon


Michal-Kolomanski:
Scope:
- [x] @AlicjaSwierad provide the icon
- [ ] click refresh, see the old EOSC icon - replace with some new icon

PR #1548, opened 2025-04-29
style: [#1546] pre-loader mod

jarekzet:
closes: #1546 

Preloader without service logo.


AlicjaSwierad (approved):

Michal-Kolomanski (changes requested):
The icon should be replaced with a new one. Its not only a deletion I believe. Please consult that with @AlicjaSwierad (plis don't approve if that is not the expected result)

Michal-Kolomanski (approved):
Nevermind, its good

Issue #1555, opened 2025-05-15
Adapters


Michal-Kolomanski:
Assumptions:
- We are not waiting for the PC to fully onboard adapters. We start mocking the data on our side. 
- Nonetheless, PC on the dev instance should onboard Adapters really soon. See here: https://dev.providers.sandbox.eosc-beyond.eu/api/swagger-ui/index.html
- After Santander, Adapters should also be available on the integration instance
https://integration.providers.sandbox.eosc-beyond.eu/api/swagger-ui/index.html

Questions: @agpul @AlicjaSwierad 
- [x] Do we want to have adapters in all collection? YES
- [x] Do we want to have adapters in the Marketplace - service detail page,  or a redirect button will be good enough? Redirect good for now, later on fetch data from Search Service API

Tickets:
Data:
- [x] Transformer - populate solr with adapters
- [x] When PC is ready - fetch data from them dynamically 

UI:
- [x] Search Service: adapters collection list view
- [x] Search Service: all collection list view with adapters
- [x] Search Service: adapters detail page view
- [ ] Search Service: guidelines detail page view with adapters
- [ ] Marketplace: service detail page view with adapters - for starting point I will do there only a redirect button with proper filters to search for adapters related to certain service. Or get that data from an search service endpoint

Example adapter:
OLD SCHEMA - look for Profile [adapters_sample_20250509 (1).json](https://github.com/user-attachments/files/20228488/adapters_sample_20250509.1.json)

Links:
- [Figma](https://www.figma.com/design/sE5vUaeTMu6jarnyD6cJXg/Adapters?node-id=2-74&p=f&t=Rx90rdd1TFRN4Y5z-0): 
- [PC schema](https://github.com/madgeek-arc/resource-catalogue/blob/develop/resource-catalogue-model/src/main/resources/schemas/json/Adapter-schema.json)
- [Profile (this is the main source of knowledge)]( https://github.com/EOSC-PLATFORM/adapter-profile/blob/main/adapter.schema.json)
- [Old profile](https://docs.google.com/spreadsheets/d/1j2xaZaAiQE-WVZJNjCYMFop3HCg0A6CdF20kqc8_HoY/edit?gid=780103951#gid=780103951)

Michal-Kolomanski:
Do not onboard adapters in Marketplace for now

Michal-Kolomanski:
@AlicjaSwierad please specify default icon for adapters

AlicjaSwierad:
[...]

PR #1597, opened 2025-06-17
feat: [#1555] Add adapters

wujuu:


Michal-Kolomanski:
@jarekzet please work not on this PR directly but create a branch from this branch. 
That will make it easier to move the code to PL DH later on

Michal-Kolomanski:
@wujuu could you please add [BETA] tag to Adapters? 
image
Also please remove [BETA] from guidelines

Michal-Kolomanski:
To fix black in backend styles, please try:
```
pipenv sync --dev
pipenv run black --preview .
```
Those 3 files are causing problems from time to time regarding styling

Michal-Kolomanski (commented):
Lets try to request for adapters logos using their "logo" property. But make sure that is bulletproof and can fast & silently fail (then it should fallback to default logo)

Michal-Kolomanski:
> Lets try to request for adapters logos using their "logo" property. But make sure that is bulletproof and can fast & silently fail (then it should fallback to default logo)

Not necessary to do so in this ticket. I created a dedicated ticket for that https://github.com/cyfronet-fid/eosc-search-service/pull/1597

Michal-Kolomanski (approved):

Issue #1650, opened 2025-08-26
Go to Search - background color


AlicjaSwierad:
Change background color to: 010B65

- [ ] beyond DH
- [ ] beyond MP

<img width="1496" height="172" alt="Image" src="https://github.com/user-attachments/assets/9e16b0bb-4ae4-452a-b50d-9f994ff4988c" />

PR #1670, opened 2025-09-10
style: [#1650] go to search topbar background

jarekzet:
close: #1650 

image

Important: need to fix also in MP


AlicjaSwierad:
Training, Libraries and IG - works 

MP and explorer do not work

AlicjaSwierad (approved):

PR #1707, opened 2025-10-20
fix: [#3561] fix user documentation

sqrasdf:
closes #3561

### DIFF
--- a/ui/apps/ui/src/app/pages/search-page/search-page.component.html
+++ b/ui/apps/ui/src/app/pages/search-page/search-page.component.html
@@ -96,7 +96,7 @@
 
 <!--          </div>-->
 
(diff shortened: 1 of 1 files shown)


ui/apps/ui/src/app/pages/search-page/search-page.component.html
```diff
@@ -96,7 +96,7 @@
 
 <!--          </div>-->
 
-          <a href="{{marketplaceUrl}}user_documentation" class="right-media-link">
```
Michal-Kolomanski: wrong link, it should be:
 https://docs.sandbox.eosc-beyond.eu/
 ticket; https://github.com/cyfronet-fid/eosc-search-service/issues/1708

PR #1709, opened 2025-10-20
fix: [#1708] fix user documentation

### LINKED ISSUE #1708: User documentation fix
Fix user documentation on the integration page

- [ ] user documentation (it should link to the corresponding page as on the sandbox: https://docs.sandbox.eosc-beyond.eu/

sqrasdf:
closes #1708

### DIFF
--- a/ui/apps/ui/src/app/pages/search-page/search-page.component.html
+++ b/ui/apps/ui/src/app/pages/search-page/search-page.component.html
@@ -96,7 +96,7 @@
 
 <!--          </div>-->
 
-          <a href="https://docs.sandbox.eosc-beyond.eu/" class="right-media-link">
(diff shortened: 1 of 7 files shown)


ui/apps/ui/src/app/pages/search-page/search-page.component.html
```diff
@@ -96,7 +96,7 @@
 
 <!--          </div>-->
 
-          <a href="{{marketplaceUrl}}user_documentation" class="right-media-link">
+          <a href="https://docs.sandbox.eosc-beyond.eu/" class="right-media-link">
```
wziajka: It would be better to create a separate variable for the documentation URL instead of hard-coding the link. This makes the code more flexible and easier to maintain across different environments

PR #1717, opened 2025-10-31
fix: [#1654] fix data source filter

### LINKED ISSUE #1654: Fix data source filter
Image

Problems:
- [ ] CESSDA is duplicated. Raw pid and pretty name
- [ ] In the dump it should be only CESSDA data source. What about those bottom values? Problem in our code or in the dump?

Scope:
- [ ] Discovery Hub
- [ ] PL Discovery Hub

Reproduce another bug:
1) click here:
Image
2) See no UI indication that some filter is selected, but we are actually filtering. Maybe dropdown data sources is just missing? https://integration.search.marketplace.sandbox.eosc-beyond.eu/search/all_collection?q=*&fq=datasource_pids:(%2221.15124%2F2shDkg%22) 

Image

sqrasdf:
closes #1654

### DIFF
--- a/ui/apps/ui/src/app/collections/data/all/filters.data.ts
+++ b/ui/apps/ui/src/app/collections/data/all/filters.data.ts
@@ -1,13 +1,14 @@
 import { IFiltersConfig } from '../../repositories/types';
 import { URL_PARAM_NAME } from './nav-config.data';
 import {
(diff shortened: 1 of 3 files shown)


Michal-Kolomanski:
squash commits into 1

PR #1722, opened 2025-11-06
chore: update dependencies

jrogoz:


backend/Pipfile
```diff
@@ -8,11 +8,10 @@ fastapi = {extras = ["all"], version = "*"}
 pydantic = "*"
 sqlalchemy = "*"
 alembic = "*"
-psycopg2 = "*"
 typer = "*"
 httpx = "*"
 asgi-lifespan = "*"
-pytest-asyncio = "*"
+pytest-asyncio = "==0.23.7"
```
Michal-Kolomanski: delete `pytest-asyncio` from here, it should be only in dev packages

agpul (approved):

PR #1902, opened 2026-06-01
Fix/fix after guideline profile changes

### LINKED ISSUE #1900: Add link to compliant services in the guideline detail page
Scope:
- [x] on the guideline's detail page - create a button which will redirect to all related services to this guideline or just list those services and make them clickable. Consult with @alicjaswieradcyfronet 

How to get related services of a given guideline?
check if we have that in the solr.
I see such a line in guidelines adapter:
relatedServices: toRelatedService(guideline.related_services ?? []),
seems like we should.

If we dont have that in solr, probably there should exists an endpoint from PC API that enables that - we need to contact them

sqrasdf:
Closes #1900

### DIFF
--- a/backend/app/settings.py
+++ b/backend/app/settings.py
@@ -103,6 +103,10 @@ class GlobalSettings(BaseSettings):
 class DevSettings(GlobalSettings):
     """Dev Settings"""
 
+    RELATED_SERVICES_ENDPOINT: Url = (
+        "https://integration.providers.sandbox.eosc-beyond.eu/api/public/interoperabilityRecord/relatedResources"
+    )
+
 
 class TestSettings(GlobalSettings):
     """Test Settings"""
--- a/ui/apps/ui/src/app/auth/auth.interceptor.ts
+++ b/ui/apps/ui/src/app/auth/auth.interceptor.ts
@@ -6,12 +6,21 @@ import {
   HttpRequest,
 } from '@angular/common/http';
 import { Observable } from 'rxjs';
+import { ConfigService } from '../services/config.service';
 
 @Injectable()
 export class AuthInterceptor implements HttpInterceptor {
-  intercept = (
+  intercept(
     req: HttpRequest<object>,
     next: HttpHandler
-  ): Observable<HttpEvent<object>> =>
-    next.handle(req.clone({ withCredentials: true }));
+  ): Observable<HttpEvent<object>> {
+    if (req.url.startsWith(ConfigService.config?.related_services_endpoint)) {
+      return next.handle(req);
+    }
+    return next.handle(
+      req.clone({
+        withCredentials: true,
+      })
+    );
+  }
 }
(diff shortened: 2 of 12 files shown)


ui/apps/ui/src/app/auth/auth.interceptor.ts
```diff
@@ -9,9 +9,21 @@ import { Observable } from 'rxjs';
...
-  intercept = (
+  intercept(
     req: HttpRequest<object>,
     next: HttpHandler
-  ): Observable<HttpEvent<object>> =>
-    next.handle(req.clone({ withCredentials: true }));
+  ): Observable<HttpEvent<object>> {
+    if (
+      req.url.startsWith(
+        'https://integration.providers.sandbox.eosc-beyond.eu/api/public/interoperabilityRecord/relatedResources/'
```
Michal-Kolomanski: This URL should be defined in the .env settings.py in the backend. Not in the frontend. 

This also will not work on production - which doesnt have integration prefix. That is way we define such variables per instance differently - not directly in the code

ui/apps/ui/src/app/pages/guidelines-page/guideline-detail-page.component.ts
```diff
@@ -54,11 +56,17 @@ export class GuidelineDetailPageComponent implements OnInit {
         untilDestroyed(this)
       )
       .subscribe((item) => {
+        console.log('Guideline loaded', item);
```
Michal-Kolomanski: debug

ui/apps/ui/src/app/pages/guidelines-page/guideline-detail-page.component.ts
```diff
@@ -54,11 +56,17 @@ export class GuidelineDetailPageComponent implements OnInit {
...
+
         this.interoperabilityGuidelineItem = { ...item } as IGuideline;
 
         this.guideline = guidelinesAdapter.adapter(
           item as Partial<IGuideline> & { id: string }
         );
+
+        console.log('Guideline id', (item as IGuideline).id);
```
Michal-Kolomanski: debug

Michal-Kolomanski (changes requested):
Please define this var in the backend - maybe already there is such variable there?

backend/app/settings.py
Michal-Kolomanski, line 83: Here is the same variable - reuse

ui/apps/ui/src/app/pages/guidelines-page/guideline-detail-page.component.ts
```diff
@@ -98,4 +101,62 @@ export class GuidelineDetailPageComponent implements OnInit {
 
     return doiRegex.test(doi) ? doi : null;
   }
+
+  private setRelatedServicesLink(id: string | number) {
```
Michal-Kolomanski, line 105: make sure that this logic hasnt been already implemented somewhere else. Are you sure that this is a new functionality?

backend/app/settings.py
Michal-Kolomanski, line 105: here integration

Michal-Kolomanski (approved):

